PR #40, opened 2014-06-13
Middlewares only touch requests inside prefix

danp:
Get middleware only touching requests within `:prefix` if one is supplied. Makes it easier to integrate with Rails and such.

### DIFF
--- a/lib/committee/middleware/base.rb
+++ b/lib/committee/middleware/base.rb
@@ -14,6 +14,16 @@ module Committee::Middleware
       @router = Committee::Router.new(@schema, options)
     end
 
+    def call(env)
+      request = Rack::Request.new(env)
+
+      if @router.includes_request?(request)
+        handle(request)
+      else
+        @app.call(request.env)
+      end
+    end
+
     private
 
     def render_error(status, id, message)
--- a/lib/committee/middleware/request_validation.rb
+++ b/lib/committee/middleware/request_validation.rb
@@ -9,19 +9,17 @@ module Committee::Middleware
       @allow_extra = options[:allow_extra]
     end
 
-    def call(env)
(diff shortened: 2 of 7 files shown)


michaelsauter:
@brandur I just came across this since I have multiple schemas in the same Rails app. Now - I looked at the code and figured that every request is unpacked, even if the request is outside the prefix. Would you be opposed to returning early from `call`, e.g. by adding this as the first line:

`return @app.call(env) unless env["REQUEST_URI"].starts_with?(@prefix)`

I can open a merge request. Actually, I was about to but then came across this issue :)

brandur:
Hey @michaelsauter, thanks for taking a look!

I might be misunderstanding something, but I think that the `RequestUnpacker` only gets run when a route is inside of the `prefix` which Committee is handling. The entry point for the `RequestValidation` middleware is in the middleware base class which switches off the path being handled by the internal router [as seen here](https://github.com/interagent/committee/blob/02b98aee68e220a9a55484745bc974f3c109376e/lib/committee/middleware/base.rb#L19-L27). The router takes `prefix` as it's being initialized.

michaelsauter:
@brandur Yikes. I somehow confused the `call` and `handle` methods in `RequestValidation` ... plus I'm using a really old version atm (1.2), which uses a different "flow" (and does unpack the request before passing it to the router). Sorry, all good then :)

brandur:
>  somehow confused the call and handle methods in RequestValidation ... plus I'm using a really old version atm (1.2), which uses a different "flow" (and does unpack the request before passing it to the router).

Hah, NP. It's been so long since the brunt of this was written that I also have to read back through everything on a fairly regular basis to understand what's going on.

> Sorry, all good then :)

Good to hear! Let us know if you see anything else.

PR #63, opened 2015-02-20
Check request.media_type instead of content_type.

dpiddy:
This will allow `Content-Type` header values such as `application/json;charset=utf-8`. Similar to https://github.com/interagent/pliny/pull/122.


brandur:
Doh. This is exactly one million times better. Thanks @dpiddy! Released as 1.6.4.

PR #64, opened 2015-02-21
Unpack complex JSON types in the request body

nwjsmith:
This allows more complex JSON types to be unpacked for validation. For example, the JSON API media type `application/vnd.api+json`


brandur:
Thanks Nate! I'm also going to patch this up to use `media_type` instead of `content_type` like seen in #63.

brandur:
Released as part of 1.7.0.

nwjsmith:
Awesome, thanks!

PR #67, opened 2015-02-24
Response Validator: match response content_type to link enc_type

swalkinshaw:
Previously the Response Validator was hardcoded to only check the response's `Content-Type` against two regexps. This meant that if you weren't using one of those, you couldn't use the response validator.

The updates are meant to match up with how things are done in the Request Validator (hence the use of `Rack::Response`) and match against a link's `encType`.

### DIFF
--- a/lib/committee/response_validator.rb
+++ b/lib/committee/response_validator.rb
@@ -11,7 +11,8 @@ module Committee
 
     def call(status, headers, data)
       unless status == 204 # 204 No Content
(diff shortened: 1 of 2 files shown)


swalkinshaw:
Note: I might have to update this to fallback to a default `encType` (`application/json`).

brandur:
@swalkinshaw Woot, this is awesome! Thanks. Sorry about the late reply here.

PR #70, opened 2015-03-16
Strip Content-Type options for response

AMekss:
Hi, After committee upgrade from 1.6.2 to 1.6.5 my tests starts to fail because of Content-Type for response is set to "application/json; charset=utf-8". Of course I can cure this easily by setting encType in my schema, but what do you think about such more general fix for that?

BR,
Arturs

### DIFF
--- a/lib/committee/response_validator.rb
+++ b/lib/committee/response_validator.rb
@@ -36,8 +36,12 @@ module Committee
 
     private
 
+    def response_media_type(response)
+      response.content_type.to_s.split(";").first.to_s
+    end
+
     def check_content_type!(response)
-      unless Rack::Mime.match?(response.content_type.to_s, @link.enc_type)
+      unless Rack::Mime.match?(response_media_type(response), @link.enc_type)
         raise Committee::InvalidResponse,
           %{"Content-Type" response header must be set to "#{@link.enc_type}".}
       end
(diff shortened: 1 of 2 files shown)


dpiddy:
Rack::Request has a `media_type` method that's similar to what you've done. We've seen need to use it elsewhere in committee recently, such as [here](https://github.com/interagent/committee/blob/7b6669bb073cff7796c9fb877fdda0989e077fe5/lib/committee/request_validator.rb#L21). How about changing to use that?

Would there ever be a need to specify content type options in encType?

AMekss:
Rack::Response doesn't have `media_type` method. How about change implementation to something like this (borrowed from the Rack::Request):
```ruby
    def response_media_type(response)
      response.content_type.to_s.split(/\s*[;,]\s*/, 2).first.to_s.downcase
    end
```
Or am I misunderstood you?

dpiddy:
My mistake! You are right. In the absence of `Rack::Response#media_type` I think borrowing the implementation from Rack::Request is good. I wonder if the rack project would take a PR to add `Rack::Response#media_type`.

AMekss:
I could try to send them PR, will come back to you

brandur:
Looks like that Rack pull might be blocked for a little while. I'm going to pull this in in the meantime to get master fixed, and release a patch level.

@AMekss Let us know if they ever get back to you! Thanks for this pull.

AMekss:
Ok, thanks

PR #75, opened 2015-04-27
Strip Content-Type options for request

dlwr:
(refs. https://github.com/interagent/committee/pull/70)

Hi,  some http clients always set Content-Type options for request, such as "application/json; charset=UTF-8".
This PR strip Content-Type options like PR#70 above.

BR,
dlwr

### DIFF
--- a/lib/committee/request_validator.rb
+++ b/lib/committee/request_validator.rb
@@ -17,9 +17,13 @@ module Committee
 
     private
 
+    def request_media_type(request)
+      request.content_type.to_s.split(";").first.to_s
+    end
+
     def check_content_type!(request, data)
-      if request.media_type && !empty_request?(request)
-        unless Rack::Mime.match?(request.media_type, @link.enc_type)
+      if request_media_type(request) && !empty_request?(request)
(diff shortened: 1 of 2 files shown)


lib/committee/request_validator.rb
```diff
@@ -17,9 +17,13 @@ def call(request, data)
...
 
+    def request_media_type(request)
+      request.content_type.to_s.split(";").first.to_s
+    end
+
     def check_content_type!(request, data)
-      if request.media_type && !empty_request?(request)
-        unless Rack::Mime.match?(request.media_type, @link.enc_type)
+      if request_media_type(request) && !empty_request?(request)
+        unless Rack::Mime.match?(request.media_type.to_s, @link.enc_type)
```
geemus: In the other pull we match against the split version of the media type, instead of the original. I think it would be good to do the same here. What do you think?

dlwr: sorry, it's very embarassing. I forgot to fix this line.

geemus:
Thanks!

PR #76, opened 2015-05-19
not enc_type but media_type to check response content-type

okitan:
Obviously, encType is for a request and mediaType is for a response.

Unfortunately there are no existing tests and test shema of specifying mediaType, and I didn't add tests about it.


geemus:
Thanks!

PR #81, opened 2015-07-23
handling optimistic_json on non-JSON body when ContentType is not `application/json`

RaVbaker:
I tried to use the optimistic_json option on non-JSON request.body and got an error because the `rescue` keyword was used in wrong way. Regardless that it is a kind of code smell: http://devblog.avdi.org/2012/11/19/rubytapas-022-inline-rescue/ inline rescue cannot have specified exception which it should catch. It always catch all `StandardError` based exceptions, as this example shows:

```ruby
2.2.1 :001 > foo
NameError: undefined local variable or method `foo' for main:Object
	from (irb):1
	from /Users/ravbaker/.rvm/rubies/ruby-2.2.1/bin/irb:11:in `<main>'
2.2.1 :002 > foo rescue NameError nil
NoMethodError: undefined method `NameError' for main:Object
	from (irb):2:in `rescue in irb_binding'
	from (irb):2
	from /Users/ravbaker/.rvm/rubies/ruby-2.2.1/bin/irb:11:in `<main>'
2.2.1 :003 > foo rescue nil
 => nil
```

Hope that this change will solve this issue.

### DIFF
--- a/lib/committee/request_unpacker.rb
+++ b/lib/committee/request_unpacker.rb
@@ -14,7 +14,11 @@ module Committee
       params = if !@request.media_type || @request.media_type =~ %r{application/.*json}
         parse_json
       elsif @optimistic_json
-        parse_json rescue MultiJson::LoadError nil
+        begin
+          parse_json
+        rescue MultiJson::LoadError
+          nil
+        end
       end
 
       params = if params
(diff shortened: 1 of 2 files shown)


brandur:
Oh man, huge +1 on inline rescues as an anti-pattern. Thanks for the contribution!

PR #82, opened 2015-07-23
Added possibility to skip content_type check in RequestValidation

### LINKED ISSUE #80: Possibility of skipping validation of request ContentType
Currently, Committee in line: https://github.com/interagent/committee/blob/2b427717cb4c6b9cd77220654c247d80b752dc48/lib/committee/request_validator.rb#L8 is verifying is request ContentType is same as the one declared in Scema `encType` field. By default JSONSchema uses `application/json` and this validator prevents POST/PUT  requests with ContentType `application/x-www-form-urlencoded`. But Rack is constructed the way that it is handling both the same way and parsing input params regardless of type of content. 

App I'm working on requires to allow for both content types, so this behavior of Rack/Sinatra is good for my case. But when I added RequestValidation from Committee it refuses requests with different ContentType than `application/json`. 

So I believe it would be good to allow skipping this `check_content_type!` method execution by setting some option when configuring this middleware (`Committee::Middleware:: RequestValidation`). 

Do you see it possible?

RaVbaker:
Small implementation to satisfy PR #80. If you don't like it, just let me know.

This fixes #80.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ Options:
 
 * `allow_form_params`: Specifies that input can alternatively be specified as `application/x-www-form-urlencoded` parameters when possible. This won't work for more complex schema validations.
 * `allow_query_params`: Specifies that query string parameters will be taken into consideration when doing validation (defaults to `true`).
+* `check_content_type`: Specifies that content_type should be verified according jsonschema definition. (defaults to `true`).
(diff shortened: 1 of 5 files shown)


brandur:
Yep. This makes sense to me. Thanks again Rafal!

brandur:
Released in 1.8.0.

PR #87, opened 2015-10-29
Allow YAML to be loaded along with JSON

brandur:
Use `YAML.load` which will load either JSON or YAML formatted schemas.
YAML is often more convenient to work with than JSON as it's much more
easily consumed by a human. Prmd already has this particular nicety, so
this changes Committee's executable to match.

We also pull the `multi_json` dependency now that JSON is already pretty
well baked into the Ruby language.

### DIFF
--- a/examples/app.rb
+++ b/examples/app.rb
@@ -4,7 +4,7 @@ require "securerandom"
 require "sinatra/base"
 
 class App < Sinatra::Base
-  SCHEMA = MultiJson.decode(File.read("schema.json"))
+  SCHEMA = JSON.parse(File.read("schema.json"))
 
   # The request validator verifies that the required input parameters (and no
   # unknown input parameters) are included with the request and that they are
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ module MyAPI
       [
         status,
         { "Content-Type" => "application/vnd.api+json" },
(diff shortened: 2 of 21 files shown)


examples/app.rb
```diff
@@ -4,7 +4,7 @@
 require "sinatra/base"
 
 class App < Sinatra::Base
-  SCHEMA = MultiJson.decode(File.read("schema.json"))
+  SCHEMA = JSON.parse(File.read("schema.json"))
```
geemus: I suppose this one could be YAML load as well? Seems fine either way. Might be good to at least call out in the README or something that either/or is acceptable as whenever I bring that up around prmd people seem surprised.

geemus:
Seems good, definitely seemed like allowing prmd to do YAML eased people in and boosted adoption overall.

examples/app.rb
brandur: Good call! Made that change in 34cb81a

brandur:
@geemus Thanks! I find it way easier in general just to generate and read in pure YAML. It sure makes any kind of manual inspection much easier anyway.

And sorry about the mergefest here! I wasn't totally sure whether anyone else was actually watching this repo ;)

geemus:
No worries, nice to see movement. Yeah, generate/read is definitely more straightforward with YAML.

PR #89, opened 2015-10-29
Create a mechanism for normalized release with Bundler's tasks

brandur:
I always forget to push commits and/or tags and/or gems during a release, so let's normalize the process using Bundler's built-in helpers for it.


brandur:
One caveat is that the format of the tag changes from "1.10.0" to one with a prefixed "v" like "v1.10.0", but that shouldn't matter all that much in the long run.

geemus:
:+1: I've always had better luck with consistency via tasks like this than my memory.

PR #91, opened 2015-12-18
Allow `committee.response` value to be replaced

brandur:
This allows a stub override to replace `committee.response` rather than
forcing it to manipulate the hash in it directly. This can be useful in
cases where an app might want to use immutable hash manipulation
functions instead of mutable ones, for example.


brandur:
Needs a rebase against master after #92 comes in.

PR #92, opened 2015-12-18
Remove Ruby 1.9.3. Add Ruby 2.0.0.

brandur:
Remove Ruby 1.9.3 from Travis and support as it's been EOLed since
February (about 10 months). This seems to be required now that the
`json_schema` gem is only compatible with 2+.

Add in Ruby 2.0.0 which is still supported (until February 2016) at
least.


brandur:
Going to pull this in, but let me know if anyone out there has any major objections to dropping 1.9. It's a tough choice, but I kind of want to keep the project modernizing at a reasonable rate. The good news is that I expect we can keep Ruby 2.0 around for a while without trouble.

geemus:
A hard choice, but agreed it is probably the old one. I am still trying to do 1.9 with fog and it is becoming progressively more painful all the time.

PR #93, opened 2016-01-16
Add Ruby 1.9.3 back to Travis test list + official 2.3

brandur:
I hate to flip flop on this, but I decided to patch json_schema to still
support 1.9.3 because it only necessitated a change in a single line. As
such, I'm bringing it back to Committee just to keep the version on life
support a bit longer (the real reason is I want to use it in
stripe-ruby's test suite and we're currently supporting versions of Ruby
well past the expiration dates right now, including 1.9.3).

Also move to 2.3.0 from 2.3.0-preview2 now that we have an official
release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,10 @@
 language: ruby
 rvm:
+  - 1.9.3
   - 2.0.0-p648
   - 2.1.8
   - 2.2.4
-  - 2.3.0-preview2
+  - 2.3.0
 
 cache: bundler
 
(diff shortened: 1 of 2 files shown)


brandur:
(And I'm still going to leave 1.9.3 out of the list in the README because I'm not too proud of it.)

geemus:
I feel your pain. Still trying to support 1.9s myself, but wish I could get out of that business...

brandur:
@geemus Hah, amen to that!

PR #95, opened 2016-04-10
Allow rack 2

mtsmfm:
Rails 5 requires rack 2

### DIFF
--- a/committee.gemspec
+++ b/committee.gemspec
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
   s.files         = Dir["{bin,lib,test}/**/*.rb"]
 
   s.add_dependency "json_schema", "~> 0.6", ">= 0.6.1"
-  s.add_dependency "rack", "~> 1.5"
+  s.add_dependency "rack", ">= 1.5"
 
   s.add_development_dependency "minitest", "~> 5.3"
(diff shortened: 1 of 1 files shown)


brandur:
Thanks!

brandur:
Released as 1.14.0.

dmathieu:
Should we really use a `>=`?
How about `['~> 1.5', '~> 2.0']`

brandur:
> How about ['~> 1.5', '~> 2.0']

@dmathieu Fair point! I guess one thing to consider is whether the fairly constrained interface to Rack that's used by Committee is ever likely to really change significantly (it's definitely possible, but based on the changes required for an upgrade from v1 to v2, we may be able to forecast that it's not probably). Thoughts?

dmathieu:
How is committee using rack?

brandur:
I think it's basically two things:

1. The `Rack::Request` interface, i.e.

    ``` ruby
    r = Rack::Request.new(env)
    r.path_info
    ```

2. The classic Rack interface of `status, headers, lines = app.call(env)` (used implicitly as Committee is middleware).

dmathieu:
Yeah, this changing in rack would require a lot of heads up.
Keeping the current global version should be fine.

brandur:
Cool :)

PR #96, opened 2016-05-13
Actually respect the validate_errors option to ResponseValidation middleware

adelcambre:
The option existed, and there was an attempt to check it. But the value passed in never made it to the validator where the option was checked.

### DIFF
--- a/lib/committee/middleware/response_validation.rb
+++ b/lib/committee/middleware/response_validation.rb
@@ -5,6 +5,8 @@ module Committee::Middleware
       @validate_errors = options[:validate_errors]
     end
 
+    attr_reader :validate_errors
+
     def handle(request)
       status, headers, response = @app.call(request.env)
 
@@ -14,7 +16,7 @@ module Committee::Middleware
           full_body << chunk
         end
         data = JSON.parse(full_body)
-        Committee::ResponseValidator.new(link).call(status, headers, data)
+        Committee::ResponseValidator.new(link, validate_errors: validate_errors).call(status, headers, data)
       end
 
       [status, headers, response]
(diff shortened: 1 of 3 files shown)


adelcambre:
I guess we can't use named args. 😢  I'll fix in a few.

lib/committee/middleware/response_validation.rb
```diff
@@ -5,6 +5,8 @@ def initialize(app, options={})
       @validate_errors = options[:validate_errors]
     end
 
+    attr_reader :validate_errors
```
brandur: Would you mind just sticking this at the top of the class (above the constructor)?

lib/committee/response_validator.rb
```diff
@@ -1,16 +1,18 @@
 module Committee
   class ResponseValidator
-    def initialize(link)
+    def initialize(link, validate_errors = false)
```
brandur: Oof, yeah, 1.9 is a huge pain. Can we change this though to use the `options = {}` pattern so that we can move it over to an optional argument easily once we drop 1.9 support? Like this:

``` ruby
    def initialize(link, options={})
      ...
      @validate_errors = options[:validate_errors]
    end
```

brandur:
Oh man, this must have gotten killed during a refactor. Thanks a lot for finding it and fixing it! I left a couple minor comments, but it's looking very good.

Also, sorry about the delay in response here. I was on vacation for a few weeks and dropped the ball on my inbox.

lib/committee/response_validator.rb
adelcambre: Yeah, good call on options, I'll get that switched over.

adelcambre:
Got those fixed up. And no problem on the delay, we didn't actually need the option, I just realized it wasn't working when I was doing some debugging. Thanks!

brandur:
Nice! Thanks again :)

Released as 1.14.1.

Issue #97, opened 2016-08-16
Validation of GET params


michaelsauter:
If I have a `GET` endpoint that takes a parameter of type integer, the validation fails, since the parameter is actually of type string. For `POST` endpoints, this problem does not exist since in the JSON payload, I can explicitly pass an integer. For GET, that's not possible. Is that true or am I missing something really obvious?

I had a look through the request validation code to see how it works but there's no immediate solution to the problem that comes to my mind. Any ideas? I guess somehow committee should allow a string that is only numbers as an integer?

michaelsauter:
Or maybe this is more related to https://github.com/brandur/json_schema? Since that is what is only allowing an `Integer` class for the integer type ...

geemus:
It may just be oversite, in our usage we pretty much only use params from body json, where they would be encoded as integers. I guess, that said, if it really is a string, maybe you should set the validation to either expect an integer or to accept both integer and string?

michaelsauter:
Hmm. The issue is that I'd like to accept integers only, but that is not possible. Committee allows me to avoid param validation in the controller, and if I have to allow string, I need to manually convert it to integers ...

So that means you also use a body json for GET requests? I think most clients would probably assume the params are added to the query string.

geemus:
I think we have largely avoided params for GET, and/or in those places where we do I think we may be failing to validate at this point (that is the state of things in some legacy places). It is a little awkward here in that json-schema is todo with validating json (which this isn't), so I could kind of go either way.

@brandur thoughts?

brandur:
> So that means you also use a body json for GET requests? I think most clients would probably assume the params are added to the query string.

That's definitely one way to go! Although Roy Fielding himself has somewhat infamously [come out against the use of request bodies on `GET` actions](http://stackoverflow.com/a/983458).

Unfortunately, there's no clear answer on this one because there's a fundamental mismatch between two encoding schemes: JSON schema is designed to validate JSON which has non-string types and can therefore unlock some more powerful features, but `GET` parameters are encoded as `application/x-www-form-urlencoded` where all values are strings.

I'd suggest one of two things:

1. In definitions for `GET` endpoints, don't use the integer type, but rather a string, and then pair it with `format` to check that the input is indeed numeric. You'll have to do some basic parsing at your application level, but at least you'll have some basic validation that your input is what you expected.
[...]

PR #99, opened 2016-09-08
Optionally coerce query params

michaelsauter:
Implementation of https://github.com/interagent/committee/issues/97#issuecomment-244687060.

I also added a note to the readme about the rules of this coercion. Not sure if this is enough and easy to understand, but it's a start.

I also had to change the link in the schema from `/apps/search` to `/search/apps` because `/apps/search` was shadowed by `/apps/{(%23%2Fdefinitions%2Fapp%2Fdefinitions%2Fname)}`. I chose to use the search example for the tests because that is also our use case - passing query values to a search endpoint, e.g. `per_page` which should be an integer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,12 +24,12 @@ Options:
 * `allow_form_params`: Specifies that input can alternatively be specified as `application/x-www-form-urlencoded` parameters when possible. This won't work for more complex schema validations.
 * `allow_query_params`: Specifies that query string parameters will be taken into consideration when doing validation (defaults to `true`).
 * `check_content_type`: Specifies that `Content-Type` should be verified according to JSON Hyper-schema definition. (defaults to `true`).
+* `coerce_query_params`: Tries to convert GET params (which are strings) into concrete types required by the schema. This works for `null` (empty value), `integer` (numeric value without decimals), `number` (numeric value) and `boolean` ("true" is converted to `true` and "false" to `false`). If coercion is not possible, the original value is passed unchanged to schema validation.
 * `error_class`: Specifies the class to use for formatting and outputting validation errors (defaults to `Committee::ValidationError`)
 * `optimistic_json`: Will attempt to parse JSON in the request body even without a `Content-Type: application/json` before falling back to other options (defaults to `false`).
 * `prefix`: Mounts the middleware to respond at a configured prefix.
 * `raise`: Raise an exception on error instead of responding with a generic error body (defaults to `false`).
 * `strict`: Puts the middleware into strict mode, meaning that paths which are not defined in the schema will be responded to with a 404 instead of being run (default to `false`).
-* `coerce_query_params`: Tries to convert GET params (which are strings) into concrete types required by the schema. This works for `null` (empty value), `integer` (numeric value without decimals), `number` (numeric value) and `boolean` ("true" is converted to `true` and "false" to `false`). If coercion is not possible, the original value is passed unchanged to schema validation.
 
 Some examples of use:
 
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ Options:
 * `prefix`: Mounts the middleware to respond at a configured prefix.
 * `raise`: Raise an exception on error instead of responding with a generic error body (defaults to `false`).
 * `strict`: Puts the middleware into strict mode, meaning that paths which are not defined in the schema will be responded to with a 404 instead of being run (default to `false`).
+* `coerce_query_params`: Tries to convert GET params (which are strings) into concrete types required by the schema. This works for `null` (empty value), `integer` (numeric value without decimals), `number` (numeric value) and `boolean` ("true" is converted to `true` and "false" to `false`). If coercion is not possible, the original value is passed unchanged to schema validation.
 
 Some examples of use:
 
--- a/lib/committee/middleware/request_validation.rb
+++ b/lib/committee/middleware/request_validation.rb
@@ -2,11 +2,12 @@ module Committee::Middleware
   class RequestValidation < Base
     def initialize(app, options={})
       super
-      @allow_form_params  = options.fetch(:allow_form_params, true)
-      @allow_query_params = options.fetch(:allow_query_params, true)
-      @check_content_type = options.fetch(:check_content_type, true)
-      @optimistic_json    = options.fetch(:optimistic_json, false)
-      @strict             = options[:strict]
+      @allow_form_params   = options.fetch(:allow_form_params, true)
+      @allow_query_params  = options.fetch(:allow_query_params, true)
+      @check_content_type  = options.fetch(:check_content_type, true)
+      @optimistic_json     = options.fetch(:optimistic_json, false)
+      @coerce_query_params = options.fetch(:coerce_query_params, false)
+      @strict              = options[:strict]
 
       # deprecated
       @allow_extra = options[:allow_extra]
--- a/lib/committee/middleware/request_validation.rb
+++ b/lib/committee/middleware/request_validation.rb
@@ -13,14 +13,13 @@ module Committee::Middleware
     end
 
     def handle(request)
-      request.env[@params_key] = Committee::RequestUnpacker.new(
-        request,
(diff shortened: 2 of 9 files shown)


michaelsauter:
Hmm, test failures are because `rack` is set to `>= 1.5`. The newest version, `2.0.1`, requires Ruby 2.2.2. Tests with Ruby version 2.2.4 and 2.3.0 have passed.

brandur:
> Hmm, test failures are because Rack dependency is set to >= 1.5. The newest Rack version, 2.0.1, requires Ruby 2.2.2. Tests with Ruby version 2.2.4 and 2.3.0 have passed.

Oh man, that's unfortunate :/ Mind restricting the Rack dependency to '< 2' for now? I don't really want to drop support for everything under 2.2.2 quite yet ...

lib/committee/middleware/request_validation.rb
```diff
@@ -2,25 +2,33 @@ module Committee::Middleware
...
       if link = @router.find_request_link(request)
+        if @coerce_query_params && !request.GET.nil? && !link.schema.nil?
+          request.env["rack.request.query_hash"].merge!(
+            Committee::QueryHashCoercer.new(
+              request.GET,
+              link.schema
+            ).call
+          )
+        end
+        request.env[@params_key] = Committee::RequestUnpacker.new(
```
brandur: So one "feature" of Committee is that an unpacked set of parameters is still available in the request's env even if no link was found, and I think we need to keep this behavior. You might have to retool so that:

1. Look for link and run query coercer if appropriate.
2. Run request unpacker.
3. Re-use link variable from (1) and run the validator.

lib/committee/query_hash_coercer.rb
```diff
@@ -0,0 +1,35 @@
...
+                coerced[k] = nil if original_val.empty?
+              when "integer"
+                coerced[k] = Integer(original_val)
+              when "number"
+                coerced[k] = Float(original_val)
+              when "boolean"
+                coerced[k] = true if original_val == "true"
+                coerced[k] = false if original_val == "false"
+              end
+            rescue ArgumentError
```
brandur: It sucks that these errors are so unspecific, but can we conditionally rescue here only if the message matches "invalid value for"? I've been bitten too many times by broad rescue statements that end up doing the wrong thing.

It might actually be a good idea to move the rescue specifically into the "integer" and "number" blocks and then match against something like "invalid value for Integer()", just for maximum certainty. The test suite should give us protection against the error message changing in the future.

lib/committee/query_hash_coercer.rb
```diff
@@ -0,0 +1,35 @@
+module Committee
+  class QueryHashCoercer
```
brandur: I've done a very poor job of this elsewhere in the library, but would you mind adding a short description on this class indicating exactly what it's for?

README.md
```diff
@@ -29,6 +29,7 @@ Options:
 * `prefix`: Mounts the middleware to respond at a configured prefix.
 * `raise`: Raise an exception on error instead of responding with a generic error body (defaults to `false`).
 * `strict`: Puts the middleware into strict mode, meaning that paths which are not defined in the schema will be responded to with a 404 instead of being run (default to `false`).
+* `coerce_query_params`: Tries to convert GET params (which are strings) into concrete types required by the schema. This works for `null` (empty value), `integer` (numeric value without decimals), `number` (numeric value) and `boolean` ("true" is converted to `true` and "false" to `false`). If coercion is not possible, the original value is passed unchanged to schema validation.
```
brandur: Looks good! Mind just keeping this list alphabetized too? Thanks!

test/query_hash_coercer_test.rb
```diff
@@ -0,0 +1,57 @@
+require_relative "test_helper"
+
+describe Committee::QueryHashCoercer do
```
brandur: Excellent new tests! Thanks :)

Would you mind just adding cases for what happens with each type in case an invalid value is encountered?

brandur:
@michaelsauter Great job! Thanks for the quick follow-up on this.

So reading this through, I think what's supposed to happen if a query parameter is an invalid type is that it fails to be converted in the coercer and then when we're merging the coercer's results in `rack.request.query_hash`, we keep the original value for that parameter, which then later fails to validate because it's of an improper type. (Correct me if I'm wrong here.) Because this relies on an interaction between components outside of the coercer itself, can we add a basic unit test (probably in `test/middleware/request_validation_test.rb`) to verify this works as expected?

lib/committee/middleware/request_validation.rb
michaelsauter: Oh okay. I did not notice that. I'll see if I can change that.

lib/committee/query_hash_coercer.rb
michaelsauter: Sure, I can do this. However, I'm not really sure if necessary in this case. I fail to see how anything in this code block could throw an ArgumentError except the conversions?
On the other hand it doesn't hurt to be extra safe and add a check ...

lib/committee/query_hash_coercer.rb
michaelsauter: Will add one.

README.md
michaelsauter: Ha, I didn't notice. Will move it further up.

test/query_hash_coercer_test.rb
michaelsauter: I'm doing that in some places even though the description doesn't reflect it. Also noticed now that I forgot to add tests for some. Will adapt.

michaelsauter:
@brandur Thanks for the quick review!

Yes, you're correct that it just merges in coerced values. So if coercion fails, the original value is going to be validated which should fail (I chose to do it that way to not have to manually raise validation errors). There's a basic integration test already in https://github.com/michaelsauter/committee/blob/a1f87d7075e11dca9f146bd6ed333aaa96725699/test/middleware/request_validation_test.rb#L97. Did you see that? Or would you like to test another case? Happy to add another one.

michaelsauter:
Regarding Rack: I can restrict it in this PR, but I'd prefer not to:

* It's not related to my changes here - I think it's better to have it in a separate PR that highlights the consequences
* It's going to prevent Committee to be installed with apps that use Rack 2. If I'm not mistaken, e.g. Rails 5 is using Rack 2, no? We're using Committee in a Rails app (though on 4.2) ...

lib/committee/query_hash_coercer.rb
brandur: >  On the other hand it doesn't hurt to be extra safe and add a check ...

Yeah, it's not so much about the code as it stands today, but rather a hedge against potential additions in the future that may be able to raise something that would be improperly caught here.

brandur:
@michaelsauter Thanks for the notes!

> Yes, you're correct that it just merges in coerced values. So if coercion fails, the original value is going to be validated which should fail (I chose to do it that way to not have to manually raise validation errors). There's a basic integration test already in https://github.com/michaelsauter/committee/blob/a1f87d7075e11dca9f146bd6ed333aaa96725699/test/middleware/request_validation_test.rb#L97. Did you see that? Or would you like to test another case? Happy to add another one.

Oh yes, I saw that one and it looks good! That said, what do you think about adding a second one of these where a query parameter is of an invalid type? I'd feel a little safer knowing that this path is properly tested.

> Regarding Rack: I can restrict it in this PR, but I'd prefer not to:

Fair enough and good point about Rails. I'll figure this out on a separate PR and we can rebase this one.

brandur:
@michaelsauter I added a patch for the Rack problem in #100 and have pulled it into master. Rebase at your leisure.

michaelsauter:
@brandur Oh, nice solution with Rack. I didn't even think of that.

I rebased and (hopefully) addressed all comments.

brandur:
Looks good to me! Thanks again @michaelsauter!

brandur:
Released as version 1.15.0.

michaelsauter:
Thanks heaps!

Updating and dropping our hacks now :)

PR #101, opened 2016-10-05
OpenAPI spec 2 support

### LINKED ISSUE #73: support for swagger json schema
Hey guys ,
just wondering if this supports [swagger json schema](https://github.com/swagger-api/swagger-spec/blob/master/schemas/v2.0/schema.json) ? 

In fact i'm thinking of a workflow where i'd first design my API with the [swagger editor](http://editor.swagger.io/#/) then extract the schema and push a mock server based on that on heroku and focus on the front end, with something like middleman... 

cc @mattolson

brandur:
Adds support for OpenAPI specification 2.

This works by moving hyper-schema specific code out into its own "driver" class and then refactoring all other classes to know how to call into the driver interface for specific functionality. A new driver for OpenAPI 2 was then added alongside the existing one for hyper-schema.

~~**Currently a WIP.** Most pieces are working, but we don't yet have parameter validation for OpenAPI 2. This is largely because it's implemented its own exotic format rather than just using JSON schemas like it does for responses. This is actually going to be changed in OpenAPI 3 (meaning that Committee will be easier to implement there), but for now we'll have to implement OpenAPI 2 parameter validation anyway.~~

Fixes #73.


dmathieu:
This looks awesome! 🌹

brandur:
> This looks awesome! 🌹

@dmathieu Thanks!

So I think I've finished most of this up (except for a re-written README which I think I'll do separately). Unfortunately, the pull request is crazy large, but any chance you want to take a quick look over to see if you can find any obvious problems? :) I tried to add quite a few new tests to make sure that we don't break anything here, but I might bring this in on a new major release anyway.

brandur:
Also cc @dpetrovics for some visibility into what I've been working on for the last couple days. I plan to re-use this along with our new OpenAPI spec to power stripe-ruby's test suite. We'll see how it goes :)

test/middleware/request_validation_test.rb
```diff
@@ -55,7 +55,7 @@ def app
 
   it "warns when sending a deprecated string" do
     mock(Committee).warn_deprecated.with_any_args
-    @app = new_rack_app(schema: File.read("./test/data/schema.json"))
+    @app = new_rack_app(schema: File.read("./test/data/hyperschema/heroku.json"))
```
dmathieu: Do we need to use the term `heroku` here? Shouldn't it be `hyper_schema`?

dmathieu: :+1:

brandur:
Alright, I think that I've got a reasonably complete implementation that I've tested with a non-trivial spec produced for the Stripe API. I plan to merge this soon (and probably bump the major version and release an alpha version that will live for a few weeks before we go all in), but I'll leave it open for a few more days in case any feedback comes in.

I've boiled the "modernized" version of the interface down to this:

``` ruby
driver = Committee::Drivers::OpenAPI.new # or Committee::Drivers::HyperSchema
schema = driver.parse(::YAML.load(file))

app = Rack::Builder.new {
  use Committee::Middleware::Stub, schema: schema

  run lambda { |_|
    [404, {}, ["Not found"]]
  }
}
```

Or for the testing module:

``` ruby
driver = Committee::Drivers::OpenAPI.new # or Committee::Drivers::HyperSchema
schema = driver.parse(::YAML.load(file))

describe MyAPI do
  include Committee::Test::Methods
  include Rack::Test::Methods

  def app
    @app
  end

  def committee_schema
    schema
  end

  it "responds according to contract" do
    get "/api"
    assert_schema_confirm
  end
end
```

The one major difference here is that I'm trying to encourage people to pre-parse a spec/schema before sending it into a middleware or testing module. The reasoning is that parsing a schema or OpenAPI spec is an expensive operations (on the order of 100s of milliseconds for a big one like Heroku or Stripe) and we don't want to be doing it on every request or test run. Passing a hash to `schema` options is still allowed, but will start to produce a warning.

All interfaces stay backward compatible (although in certain cases may produce new warnings as mentioned above).

brandur:
Released as 2.0.0pre.

For Heroku users and others: after working with OpenAPI 2.0 for the last few days, I'd probably recommend _not_ switching it for now. Although a few things are better than hyper-schema, it's got a few problems. For example, it's not possible to use non-scalar types for parameters or have endpoints that might return more than one type.

Most of these limitations will be fixed for OpenAPI 3.0, and its release shouldn't be too far off (although I'm not sure exactly when it can be expected). If you're going to migrate, you'd probably save yourself some time by waiting for the next version.

geemus:
@brandur Thanks for the great work here and your insights, I was wondering how ready it was for adoption. Sounds like maybe the wait shouldn't be too long at least.

brandur:
> Thanks for the great work here and your insights, I was wondering how ready it was for adoption. Sounds like maybe the wait shouldn't be too long at least.

@geemus NP! I'll try to keep on top of that OpenAPI 3.0. Once that's through, it would be awesome if we could consider having Heroku/Lugg/Citus/Stripe/etc. re-convergence on that as a standard.

dmathieu:
🎉

PR #103, opened 2016-10-13
Support for body parameters in OpenAPI 2.0

brandur:
Adds support for body parameters [1] in OpenAPI 2.0. These is a special
parameter type that is allowed to describe itself with a JSON schema
with the caveat that it cannot be mixed with form parameters.

Here we take a similar approach to parsing response schemas and optimize
by accumulating all body parameters into a single synthetic schema as we
parse our routes and then parsing them all at once as a last step.

[1] https://github.com/OAI/OpenAPI-Specification/blob/39624ac7ee6960a5540e71a6b17a15f53b20048d/versions/2.0.md#parameterObject

### DIFF
--- a/lib/committee/drivers/open_api_2.rb
+++ b/lib/committee/drivers/open_api_2.rb
@@ -89,43 +89,54 @@ module Committee::Drivers
         self.link_data = link_data
       end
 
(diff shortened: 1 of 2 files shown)


brandur:
This works so far in my tests so I'm going to pull it in and cut another prerelease.

PR #104, opened 2016-10-15
Option for coercing form parameters

brandur:
Adds an option for type coercion for parameters encoded with
`application/x-www-form-urlencoded` like we already have for path and
query parameters. This allows specs based off this non-JSON encoding to
successfully validate against a schema. The new option is enabled by
default for OpenAPI 2.0 but not for hyper-schema for reasons of backward
compatiblity.


brandur:
Released as 2.0.0.pre3.

PR #105, opened 2016-10-16
Convert reqest parameter by format

ota42y:
The string which set date-time format in JSON scheme, this gem check type and format.
But, even if the parameter is correct, it's the string object in the application code.
So we need convert string to date-time.

In the GET request, the parameter convert string to other type such like integer, float, boolean and It's very useful.
But, we don't care the string format.

So when the string pass the type and format check, we should convert string to datetime object.
This feature is very useful because the JSON scheme writer specify the date-time parameter.
# TODO
- [x] write a test
- [x] add option
  - If user don't want to use, disable this feature
  - I think this feature default enable
- [x] add README


brandur:
@ota42y Thanks for the pull!

So I'm not totally convinced that trying to coerce date-time strings into time objects is something that we want to handled at Committee's layer here. Coercion in some cases makes sense, for example when you're trying to change an untyped set of parameters like one from a query string into something that can be validated by JSON schema, but remember in this case the schema is promising that a date-time is foremost a _string_ type and secondarily one with a certain format that looks like a date-time.

Furthermore, just scanning [the list of other built-in formats](http://json-schema.org/latest/json-schema-validation.html#anchor107), I don't see anything else that we'd be likely to re-use this for.

Note that the library will already _validate_ that a date-time field is something that it recognizes (i.e. a string encoded as RFC 3339). That means that after the request validator has run, you should be able to easily parse those strings from your app without fear of invalid values.

Thoughts?

qsona:
@brandur 

Hi, I'm co-worker with @ota42y . I want to add some information.

## our motivation

We're trying to migrate from [grape](https://github.com/ruby-grape/grape) and grape-swagger to prmd and committee.
By using grape we can also declare request params, so we can

- validate requests
- coerce type of request params
- generate api documentation of request params (grape-swagger)
- run mock server

But, grape doesn't care about response. This is why we want to use JSON Hyper Schema,  and prmd / committee as the implementations of it. Now, we can also

- validate requests (committee)
- validate response (committee)
- coerce type of request params (committee)
- generate api documentation of request params and response body (prmd)
- run mock server (committee)

and our goal is almost achieved.

Just one thing remained we want to gain is type coercing. For example, in grape, it can be written as follows:

```rb
params do
  requires :date, type: Date
  requires :int, type: Integer
end
```

and, (in Rails form, ) `params[:date]` is automatically coerced and it's a Date object. `params[:int]` is an Integer.

I want to do this by using committee. For primitive types, `coerce_query_params` and `coerce_form_params` options work well. But, option for non-primitive types (e.g. Date) does not exist.

We focus to `format` of json schema specification.

```
"type": "string",
"format": "date-time"
```

It has some built-in formats but end-user can add any formats. By using it, we can get params as customized ruby objects.
It can reduce much of application code of validating in controller, such as follows

```
begin
  date = Date.parse(params[:date])
rescue ArgumentError
  raise ActionController::BadRequest
end
```

because `params[:date]` is already a Date object.

## design

Currently I am thinking following plans.

- Of cource, it should be optional.
- Committee prepares built-in coerce function for built-in formats
  - e.g. format 'date-time', coercing to Time Object
- User can define format type, and coerce function
  - e.g. format 'date', coercing to Date Object

We can contribute this project if we can agree about the design. Thanks.

qsona:
@brandur Perhaps my previous post was not a clear answer to your post.

My current idea is that checking type and coercing type should be atomic.

As you said, params can be parsed without error check if they are already passed through the validator. Even so, the detail logic of checking parameter is out of my concerns, so I'm a bit worried when coercing to some object.

I'm aware the question that if it's really necessary for committee's layer. For example, a parameter of date-time field may sometimes be wanted to coerce to Time object, but sometimes to DateTime object, but it cannot be specified individually because json schema can't have those options.

But I still think that it's quite convenient if application layer doesn't need to concern and can get Date or Time object directly.

brandur:
Hey @qsona,

Thanks for the detailed write-up. Thinking about this a little more, I think I'd agree that date-times are probably enough of a special case that it would make sense to optionally coerce them. Really it would be nice if they were part of JSON it self so we wouldn't have to ask ourselves these sorts of questions.

However, could you change this patch a little so that an explicitly named parameter turns the behavior on (e.g. `coerce_date_times`)? We'll also have to leave it off by default so as to better maintain backward compatibility.

Thanks!

qsona:
@brandur Sorry my response is late. I agree with whole your opinion. We will send a pull request later. Thank you.

ota42y:
@brandur 
I'm sorry  I was so late.
I add coerce_date_times option and  test :)

README.md
```diff
@@ -30,6 +30,7 @@ Options:
 * `prefix`: Mounts the middleware to respond at a configured prefix.
 * `raise`: Raise an exception on error instead of responding with a generic error body (defaults to `false`).
 * `strict`: Puts the middleware into strict mode, meaning that paths which are not defined in the schema will be responded to with a 404 instead of being run (default to `false`).
+* `coerce_date_times`: Convert the string with `"format": "date-time"` parameter to DateTime object (default to `false`).
```
brandur: Sorry for the nit, but would you mind just putting this in alphabetical order?

lib/committee/middleware/request_validation.rb
```diff
@@ -58,6 +59,15 @@ def handle(request)
       if link
         validator = Committee::RequestValidator.new(link, check_content_type: @check_content_type)
         validator.call(request, request.env[@params_key])
+
+        if @coerce_date_times
+          request.env["rack.request.query_hash"].merge!(
+              Committee::ParameterConverter
+                  .new(request.env["rack.request.query_hash"], link.schema)
```
brandur: Do you think we could retool this interface slightly?

`ParameterConverter` is quite a generally named class that seems like it could be doing work beyond just coercing datetimes. Instead of checking for `coerce_date_times` here, can we pass that into this class' constructor and have it make that decision? Like this:

``` ruby
request.env["rack.request.query_hash"].merge!(
    Committee::ParameterConverter
        .new(request.env["rack.request.query_hash"], link.schema,
          :coerce_date_times => @coerce_date_times)
```

Thoughts?

lib/committee/parameter_converter.rb
```diff
@@ -0,0 +1,44 @@
+module Committee
+  class ParameterConverter
```
brandur: Would you mind if we renamed this to `ParameterCoercer` and the method `#convert!` to `#call`?

lib/committee/parameter_converter.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    def convert!
+      coerced = {}
+      @schema.properties.each do |k, s|
+        original_val = @params[k]
+
+        unless original_val.nil?
+          s.type.each do |to_type|
+            case to_type
+            when "string"
```
brandur: Would you mind if we moved both the check on type and format into the same place? I think it's a little clearer as to what's going on. i.e.

``` ruby
if to_type == "string" && s.format == "date-time"
  coerced_val = parse_date_time(original_val)
```

lib/committee/parameter_converter.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    private
+
+      def coerce_string_val(original_val, scheme)
+        case scheme.format
+          when "date-time"
+            begin
+              DateTime.parse(original_val)
+            rescue ArgumentError => e
```
brandur: What happens to this exception if it's thrown from within the `RequestValidation` middleware? We should make sure that it doesn't manifest as a 500. It might be worth adding in a test for that case in `test/middleware/request_validation_test.rb`.

test/middleware/request_validation_test.rb
```diff
@@ -17,6 +17,16 @@ def app
...
 
+  it "pass datetime string and coerce_date_times option enable" do
+    @app = new_rack_app(coerce_date_times: true, schema: hyper_schema)
+    params = {
+        "update_time" => "2016-04-01T16:00:00.000+09:00"
+    }
+    header "Content-Type", "application/json"
+    post "/apps", JSON.generate(params)
+    assert_equal 200, last_response.status
+  end
```
brandur: Excellent! Thanks for adding the broad test here and the more specific unit tests for `ParameterConverter` as well.

ota42y:
I'm fixed :)

brandur:
@ota42y Sweet. Thank you!

brandur:
I've released this as 2.0.0.pre4 for the time being. I'm hoping to get the full 2.0.0 release out relatively shortly ...

PR #111, opened 2017-01-16
:bug: if no schema api, error occured in ParameterCoercer

ota42y:
I think we should write JSON Schema for all API :smiling_imp:
But, unfortunately there are bad people who do not write JSON Schema.
So few APIs get errors when use `coerce_date_times` . :scream:

I fix this error 😇

### DIFF
--- a/lib/committee/parameter_coercer.rb
+++ b/lib/committee/parameter_coercer.rb
@@ -9,6 +9,8 @@ module Committee
 
     def call
       coerced = {}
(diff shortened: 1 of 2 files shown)


brandur:
>  But, unfortunately there are bad people who do not write JSON Schema.

LOL! So true, haha!

Thanks for the fix!

brandur:
Released as 2.0.0.pre5.

PR #113, opened 2017-02-01
Embed response schema in env + more generators

brandur:
This one is a bit of a grab bag, but has two major goals:

* Embeds the JSON schema of a generated response in the new env key
  `committee.response_schema`. This is useful so that downstream
  applications can tweak the generated response based off of it.

* Adds a few new simple data generators: return an empty array for an
  array type, and return the first enum value for a schema with an enum.


brandur:
Released as 2.0.0.pre6.

PR #115, opened 2017-02-22
Use `assert_nil` to avoid deprecation message

qsona:
Some deprecation message are shown when running tests, so I fixed it.
See: https://github.com/seattlerb/minitest/pull/626
Personally I really don't like this change (I believe unit test API must be consistent), but it seems to be a determined policy in minitest.


brandur:
Works for me. Thanks @qsona!

PR #116, opened 2017-02-22
Add cases for coercing to boolean values

### LINKED ISSUE #110: Suggestion: behavior of StringParamsCoercer for boolean
The current behavior that how StringParamsCoercer coerces string to boolean is:
https://github.com/interagent/committee/blob/a99f1b4f1eda0eb9c269efa6d55e451f5d1d3b81/lib/committee/string_params_coercer.rb#L39-L42

so, the `s` value is regarded as boolean, it coerces (or doesn't coerce) query such as follows.

```
?s=true #=> true
?s=false #=> false
?s=1 #=> "1"
?s= #=> ""
```

After the value passed through coercer, it is sometimes boolean but sometimes not. So application code needs more validation code and it often becomes a bit complicated.

As far as I know, there is no rule that how the form-urlencoded param should be coerced to boolean value,
but in my experience, `s=` is often regarded as `s: false`, and `s=1` as `s: true`. (Of course it's debatable. )

So, here is my suggestion: if the string value is empty or "false", it is coerced to false. else, it is coerced to true.

```
?s=1 #=> true
?s=asdf #=> true
?s= #=> false
```

Any thoughts? Thanks.

qsona:
Partially fixed #110 .

### DIFF
--- a/lib/committee/string_params_coercer.rb
+++ b/lib/committee/string_params_coercer.rb
@@ -37,8 +37,8 @@ module Committee
                 raise e unless e.message =~ /invalid value for Float/
               end
             when "boolean"
-              coerced[k] = true if original_val == "true"
-              coerced[k] = false if original_val == "false"
+              coerced[k] = true if original_val == "true" || original_val == "1"
+              coerced[k] = false if original_val == "false" || original_val == "0"
             end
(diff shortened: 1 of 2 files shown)


brandur:
Nice!

PR #118, opened 2017-03-16
Show NotFound error message with :raise option

k0kubun:
In development environment, :raise option is enabled and I couldn't see a kindful error message. I want to see that on development too.


brandur:
Good point.

brandur:
Released for now as 2.0.0.pre7.

Thanks for the patch!

PR #119, opened 2017-05-10
Support recursive coerce

ota42y:
`coerce_date_times` option didn't work recursive object and array object so I did fix.

For example, the commitee didn't convert string object to datetime object when `coerce_date_times` is true.

```
{
  "array_property" => [
  {
    "update_time" => "2016-04-01T16:00:00.000+09:00",
  }
}
```

ParameterCoercer convert top level(?) keys only.
So I implement recursive convert.

we need fix when this option enable
- [x] coerce_query_params
- [x] coerce_form_params

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ Options:
 * `allow_query_params`: Specifies that query string parameters will be taken into consideration when doing validation (defaults to `true`).
 * `check_content_type`: Specifies that `Content-Type` should be verified according to JSON Hyper-schema definition. (defaults to `true`).
 * `coerce_date_times`: Convert the string with `"format": "date-time"` parameter to DateTime object (default to `false`).
+* `coerce_recursive`: Convert data in array object and nested object (default to `false`).
 * `coerce_form_params`: Tries to convert POST data encoded into an `application/x-www-form-urlencoded` body (where values are all strings) into concrete types required by the schema. This works for `null` (empty value), `integer` (numeric value without decimals), `number` (numeric value) and `boolean` ("true" is converted to `true` and "false" to `false`). If coercion is not possible, the original value is passed unchanged to schema validation.
 * `coerce_query_params`: The same as `coerce_form_params`, but tries to coerce `GET` parameters encoded in a request's query string.
 * `coerce_path_params`: The same as `coerce_form_params`, but tries to coerce parameters encoded in a request's URL path.
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Options:
 * `allow_query_params`: Specifies that query string parameters will be taken into consideration when doing validation (defaults to `true`).
 * `check_content_type`: Specifies that `Content-Type` should be verified according to JSON Hyper-schema definition. (defaults to `true`).
 * `coerce_date_times`: Convert the string with `"format": "date-time"` parameter to DateTime object (default to `false`).
-* `coerce_recursive`: Convert data in array object and nested object (default to `false`).
+* `coerce_recursive`: Convert data in array object and nested object (default to `true`).
 * `coerce_form_params`: Tries to convert POST data encoded into an `application/x-www-form-urlencoded` body (where values are all strings) into concrete types required by the schema. This works for `null` (empty value), `integer` (numeric value without decimals), `number` (numeric value) and `boolean` ("true" is converted to `true` and "false" to `false`). If coercion is not possible, the original value is passed unchanged to schema validation.
 * `coerce_query_params`: The same as `coerce_form_params`, but tries to coerce `GET` parameters encoded in a request's query string.
 * `coerce_path_params`: The same as `coerce_form_params`, but tries to coerce parameters encoded in a request's URL path.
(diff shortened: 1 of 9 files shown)


brandur:
@ota42y Let me know when you want a review on this one!

ota42y:
Sorry, I forgot this... 😭 

And I found a bug in this PR, so I'll change `Committee::ParameterCoercer#call` and `Committee::StringParamsCoercer#call` to mutable method from immutable method.

Now `Committee::ParameterCoercer#call` don't change original object.
This class create new hash object and merge original hash in `Committee::Middleware::RequestValidation`
So I implement recursive function as immutable method.
This method create new Hash or Array object and copy data.

But when user convert request parameter to Hash like object (e.g. ActiveSupport::HashWithIndifferentAccess) before committee convert, my function change it to normal Hash object.

like this.
(This is sample code. In the actual case, the user uses rack middleware)

```ruby
class HashLikeObject < Hash; end

recursive_object = HashLikeObject.new
recursive_object['datetime'] = '2016-04-01 16:00:00'

params = {
  data: recursive_object
}

new_params = Committee::ParameterCoercer.
            new(recursive_object, link.schema, coerce_date_times: @coerce_date_times).
            call!

new_params[:data].class == params[:data].class # Hash == HashLikeObject so it's false
```

`Committee::StringParamsCoercer` have same problem so I'll change too.

ota42y:
I'm sorry to keep you waiting but it's finally completed :)

ota42y:
@brandur I'm sorry to keep you waiting but it's finally completed :)

callmegema:
@brandur
I've tested this feature at my project already. I think it's working correctly.

lib/committee/parameter_coercer.rb
```diff
@@ -5,31 +5,63 @@ def initialize(params, schema, options = {})
...
-      return coerced unless @schema.respond_to?(:properties)
+    def call!
+      coercer_object!(@params, @schema)
+    end
+
+    private
 
-      @schema.properties.each do |k, s|
-        original_val = @params[k]
+      def coercer_object!(hash, schema)
```
brandur: Nice use of recursion all over here!

Do you think that we could rename each of these methods to `coerce_*` instead of `coercer_*`?

lib/committee/parameter_coercer.rb
```diff
@@ -5,31 +5,63 @@ def initialize(params, schema, options = {})
       @schema = schema
 
       @coerce_date_times = options.fetch(:coerce_date_times, false)
+      @coerce_recursive = options.fetch(:coerce_recursive, false)
     end
 
-    def call
-      coerced = {}
-      return coerced unless @schema.respond_to?(:properties)
+    def call!
```
brandur: Thanks for renaming this to use a bang.

lib/committee/string_params_coercer.rb
```diff
@@ -10,41 +10,90 @@ module Committee
...
+      coercer_object!(@query_hash, @schema)
     end
 
-    def call
-      coerced = {}
-      @schema.properties.each do |k, s|
-        original_val = @query_hash[k]
+    private
+
+      def coercer_object!(hash, schema)
```
brandur: And same with these please. `coercer_*` -> `coerce_*`.

README.md
```diff
@@ -25,6 +25,7 @@ Options:
 * `allow_query_params`: Specifies that query string parameters will be taken into consideration when doing validation (defaults to `true`).
 * `check_content_type`: Specifies that `Content-Type` should be verified according to JSON Hyper-schema definition. (defaults to `true`).
 * `coerce_date_times`: Convert the string with `"format": "date-time"` parameter to DateTime object (default to `false`).
+* `coerce_recursive`: Convert data in array object and nested object (default to `false`).
```
brandur: Just OOC, is the extra option for backwards compatibility or something? It seems to me that having coercion that'll recurse into your more complex objects is probably the default that most users would want.

brandur:
> I've tested this feature at my project already. I think it's working correctly.

@callmegema Thanks for letting me know! Finally got around to leaving a review here.

README.md
ota42y: Yes, this option for backwards compatibility...
However, 2.0.0 is major version up, so I think it is okay to default true.
What do you think?

(In other words, we want to consider the default true for other options as well)

brandur: Good thinking on the backwards compatibility!

Yes, lets ship this as part of 2.0.0 though. I think it's best as a default, and we're in the process of a major version bump anyway. I'll make sure to note it conspicuously in the CHANGELOG. Can you make the change? Thanks!

brandur:
@ota42y Was aebef54 possibly committed by accident?

ota42y:
@brandur Sorry, I fixed this commit 🙇

brandur:
Thanks @ota42y! I think we're getting pretty close, but I think just two more tests need to be updated for the default recursive coercion (e.g. "passes given a nested datetime and with coerce_recursive=false and coerce_date_times=true on POST endpoint"). Mind taking a quick look at the best suite?

ota42y:
Surely that is a good idea.
I already wrote a test to check the default value, so I modified it further.

brandur:
Thanks @ota42y! I hope you don't mind, but I'm going to make a couple tweaks to documentation before release.

brandur:
Released as 2.0.0!!

Just FYI, I moved the upgrade notes out of the README so that they're [in the CHANGELOG file only](https://github.com/interagent/committee/blob/master/CHANGELOG) given that there's been quite a few changes going up to 2.0.0.

PR #122, opened 2017-11-07
Validate header parameters

tzmfreedom:
On current version of committee, parameters in header is not validated when parameter is defined in header on OpenAPI 2.0 format.

So the following OpenAPI description is not working because current version does't extract and validate header values.
```yaml 
...
properties:
  name:
    type: string
    in: header
...
```

This pull request solve them.

* If allow_header_params is enabled, header values is extracted as parameter.
* On OpenAPI driver, header parameter is uppercased because we treat header key as CGI header values, that is uppercased and converted hyphen to underscore.

### DIFF
--- a/lib/committee/drivers/open_api_2.rb
+++ b/lib/committee/drivers/open_api_2.rb
@@ -81,19 +81,66 @@ module Committee::Drivers
       # data.
       attr_accessor :target_schema
 
+      attr_accessor :header_schema
+
       def rel
         raise "Committee: rel not implemented for OpenAPI"
+      end
+    end
+
+    class SchemaBuilder
+      def initialize(link_data)
+        self.link_data = link_data
+      end
+
+      private
+
+      LINK_REQUIRED_FIELDS = [
+        :name
+      ].map(&:to_s).freeze
+
+      attr_accessor :link_data
+
+      def check_required_fields!(param_data)
+        LINK_REQUIRED_FIELDS.each do |field|
+          if !param_data[field]
+            raise ArgumentError,
+                  "Committee: no #{field} section in link data."
+          end
+        end
+      end
+    end
+
+    class HeaderSchemaBuilder < SchemaBuilder
+      def call
+        if link_data["parameters"]
+          link_schema = JsonSchema::Schema.new
+          link_schema.properties = {}
+          link_schema.required = []
+
+          header_parameters = link_data["parameters"].select { |param_data| param_data["in"] == "header" }
+          header_parameters.each do |param_data|
+            check_required_fields!(param_data)
+
+            param_schema = JsonSchema::Schema.new
+
+            param_schema.type = [param_data["type"]]
+
+            link_schema.properties[param_data["name"]] = param_schema
+            if param_data["required"] == true
+              link_schema.required << param_data["name"]
+            end
+          end
+
+          link_schema
+        end
       end
     end
 
     # ParameterSchemaBuilder converts OpenAPI 2 link parameters, which are not
     # quite JSON schemas (but will be in OpenAPI 3) into synthetic schemas that
     # we can use to do some basic request validation.
-    class ParameterSchemaBuilder
-      def initialize(link_data)
-        self.link_data = link_data
-      end
-
+    class ParameterSchemaBuilder < SchemaBuilder
       # Returns a tuple of (schema, schema_data) where only one of the two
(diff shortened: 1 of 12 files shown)


lib/committee/drivers/open_api_2.rb
```diff
@@ -133,6 +133,9 @@ def call
                 param_schema.items = param_data["items"]
               end
 
+              # if parameter is defined in header, uppercase name parameter to recognize CGI header.
+              param_data["name"].upcase! if param_data["in"] == "header"
```
brandur: Can we just perform the `upcase` when doing the validation? I think having all the logic in one place is just going to make everything in a little more readable.

lib/committee/request_unpacker.rb
```diff
@@ -3,11 +3,12 @@ class RequestUnpacker
     def initialize(request, options={})
       @request = request
 
-      @allow_form_params  = options[:allow_form_params]
-      @allow_query_params = options[:allow_query_params]
-      @coerce_form_params = options[:coerce_form_params]
-      @optimistic_json    = options[:optimistic_json]
-      @schema             = options[:schema]
+      @allow_form_params   = options[:allow_form_params]
```
brandur: What do you think about having `allow_header_params` on by default? As you noticed, it seems to be part of the spec, and the reason that `allow_query_params` and `allow_form_params` exist is that they're doing something that's a little abnormal from the standard Committee conventions. Header parameters are almost entirely orthogonal to other types of parameters, and I suspect it's fine just to always allow them.

lib/committee/request_unpacker.rb
```diff
@@ -87,5 +90,14 @@ def parse_json
...
     end
+
+    def headers
+      env = @request.env
+      env.keys.grep(/HTTP_/).inject({}) do |headers, key|
+        headerized_key = key.gsub(/^HTTP_/, '').gsub(/_/, '-')
+        headers[headerized_key] = env[key]
+        headers
+      end
+    end
```
brandur: I'm a tad bit concerned about the potential for different types of parameters to start colliding. For example, you could have a param called `Encoding` that appeared as both a header and as a form parameter, and Committee would have a hard time differentiating between the two.

What do you think about separating headers out into separate buckets? Maybe the best solution would be if `RequestUnpacker` returned a `[headers, params]` where each element could be validated separately, but a cheaper solution might be to just have a special `__committee_headers` key in the `params` hash for now that's extracted during validation. Thoughts?

tzmfreedom: As you pointed out, I think header and parameter are orthogonal too.
In current PR implementation, collision is occured and this is not good for us.
So separating headers out into separate buckets is good idea.

I think the best solution is separeting params and headers on RequestUnpacker#call too.

The following is idea for implementation.

* `Committee::Drivers::OpenAPI2::Link` has new `header_schema` attribute that is `JsonSchema::Schema` Object to validate header.
  * `Committee::Drivers::HyperSchema` does't have `header_schema`. This attribute only be in OpenAPI.
* `RequestUnpacker` returns array, params and headers.
* `RequestValidator` validates params and headers separately. params are validated with Link#schema, headers are validated with Link#header_schema.

If the above solution looks good, I'll implement it.

lib/committee/request_unpacker.rb
tzmfreedom: Thank you for your pointing it out.
I think `allow_header_params` should be true by default too. 
In the same way as other parameter, I'll set default parameter on `RequestValidation`'s constructor.

lib/committee/request_unpacker.rb
brandur: Thanks for thinking it through so thoroughly! Yes, that sounds good to me.

tzmfreedom:
I revised.
https://github.com/interagent/committee/pull/122/commits/2eeec414655778426a153df505272419f2d72b4f

And, I fixed some test because of redundant variable assignment.
https://github.com/interagent/committee/pull/122/commits/719db3952a04b5b85b9c7d50d98ad59961cb8d23

Please review it and let me know if there are any comments.

brandur:
Thank you for the significant chunk of work here @tzmfreedom! Once again, apologies for the delay in getting back to you on it — if it happens again in the future, feel free to bump the thread to get it back to the top of my inbox.

This looks good me!

brandur:
Released as 2.1.0.

PR #124, opened 2018-01-25
support new ruby versions

ota42y:
add newest ruby versions


brandur:
Looks good to me. Thanks!

PR #127, opened 2018-04-06
bundled executable file path does not match installed script file path

hokuma:
Is executable file path check necessary?

Because of this check, below command does nothing.

```
$ gem install committee
$ committee-stub -h
# nothing to happen
```

MyEnvironment:

```
~ $ rbenv -v                                                                                                                                                                                               00:44:37
rbenv 1.1.1-gc8ba27f
~ $ ruby -v                                                                                                                                                                                                00:44:53
ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin15]
```

After inserting debug print in `bin/committee-stub` and executing it, the resuls is as follows

```
p $0
=>"MY_HOME/.rbenv/versions/2.4.1/bin/committee-stub"
p __FILE__
=>"MY_HOME/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/committee-2.1.0/bin/committee-stub"
```

### DIFF
--- a/bin/committee-stub
+++ b/bin/committee-stub
@@ -13,10 +13,9 @@ parser.parse!(args)
 
 if options[:help] || !options[:driver]
   puts parser.to_s
-  exit
+else
+  driver = Committee::Drivers.driver_from_name(options[:driver])
+  schema = driver.parse(::YAML.load(File.read(args[0])))
+
+  Rack::Server.start(app: bin.get_app(schema, options), Port: options[:port])
 end
-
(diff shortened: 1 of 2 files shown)


brandur:
(Sincere apologies for this epic delay.)

Absolutely right — I don't understand how I never caught this before. I guess because Committee is usually just being bundled into projects. Thank you!

brandur:
Released as 2.1.1.

PR #128, opened 2018-07-20
Support OpenAPI 3.0

### LINKED ISSUE #126: Does committee gem have plans to support OpenAPI 3.0?
Does committee gem have plans to support OpenAPI 3.0?

I would be happy if you would consider it.

KentaYoshitani:
fix #126

## Problem

- gem json_schema does not support `components` as a key for parameter schema and raise error  `NoMethodError: Schema does not respond to #components` if key is not `definitions`

## Temporary measure

- convert `components` key into `definitions` key by `convert_components_to_definitions`


KentaYoshitani:
@brandur Is it correct not to support `components` key as a JsonSchema specification? I'm sorry but I couldn't find a solution even after reading spec. Can you comment on this PR?

KentaYoshitani:
The reason why failing test is `Coverage (99.77%) is below the expected minimum coverage (100.00%).
` and this failed before my commit.
I am gonna fix this PR later!

yhirano55:
> The reason why failing test is Coverage (99.77%) is below the expected minimum coverage (100.00%). and this failed before my commit.
> I am gonna fix this PR later!

@KentaYoshitani Thanks for your reporting. Currently it has fixed.

KentaYoshitani:
@yhirano55 thx!

lib/committee/drivers/open_api_3.rb
```diff
@@ -0,0 +1,372 @@
...
+    ].map(&:to_s).freeze
+
+    def find_best_fit_response(link_data)
+      response_data_200 =
+        link_data["responses"]['200'] || link_data["responses"][200]
+
+      response_data_201 =
+        link_data["responses"]['201'] || link_data["responses"][201]
+
+      if response_data_200 && digged_response_data_200 = response_data_200.dig("content", "application/json", "schema")
```
yhirano55: @KentaYoshitani `Hash#dig` is available from Ruby 2.3 or higher.

lib/committee/drivers/open_api_3.rb
```diff
@@ -0,0 +1,381 @@
...
+    def parse(data)
+      data = convert_components_to_definitions(data)
+      REQUIRED_FIELDS.each do |field|
+        if !data[field]
+          raise ArgumentError, "Committee: no #{field} section in spec data."
+        end
+      end
+
+      openapi_version = Gem::Version.create(data['openapi'])
+      unless Gem::Version.create('3.0.0') <= openapi_version && openapi_version < Gem::Version.create('4.0.0')
```
ota42y: I think we don't need to check the upper version.

Because OpenAPI 4 has not been released, so we don't know it's breaking changes or not.
If there is breaking changes in v4 the user can use v3, but if there isn't breaking changes, the user can't use v4.

I think that you may allow it until problems occurred.

austinh:
Hello wanted to check in on this PR. Would love to use this for my app where I am using OpenAPI 3.0. This gem is killer for BDD and Contract First Development :)

KentaYoshitani:
I'm gonna close this PR because,
1, New PR (https://github.com/interagent/committee/pull/145/files) which is not depend on json_schema gem is now WIP.
2, I could not find solution of the problem written on comment above.

New PR is made by @ota42y  and he is my colleague, so we can discuss about OpenAPI3 committee anytime. I also talked with him about closing this PR.
I wish new PR will be merged and works well.

PR #131, opened 2018-08-04
Fix Travis Configuration

yhirano55:
# Summary

* According to README, this gem is not supporting v1.9.3. So I've deleted 1.9.3 from CI configuration.
* Fix rvm versions:
  * Use 2.5.1 instead of 2.5.0
  * Use 2.4.4 instead of 2.4.3
  * Use 2.3.7 instead of 2.3.6
  * Use 2.2.10 instead of 2.2.9
* Add 2.5 as supporting versions to README


yhirano55:
@brandur Review this please. If it merges this, it'll fix build error in CI :tada:

brandur:
Thanks @yhirano55!

PR #133, opened 2018-08-05
Add required_ruby_version to gemspec

yhirano55:
## Summary

* This gem supports ruby version 2.0.0 or higher.
* It seems it's better to describe required ruby version in its gemspec.


brandur:
Looks good to me. Thanks!

PR #136, opened 2018-08-05
Update development dependencies

yhirano55:
## Summary

* Update development dependencies (rake)
* Suppress warning ambiguous first argument.


yhirano55:
Currently, should we support ruby v2.0 or v2.1?

brandur:
> Currently, should we support ruby v2.0 or v2.1? How long are we planning to support Ruby v2.0 or v2.1?

I usually try to support old versions as long as possible until supporting them starts to cause a lot of pain trying to develop other things, and then I drop them. There isn't really a strong policy around it right now though.

yhirano55:
@brandur I agree with you. Thanks!

PR #146, opened 2018-10-30
drop support for old ruby version

ota42y:
ruby 2.2 and older is EOL.

Now I create OpenAPI3 feature and related gem require newer ruby version.
https://github.com/interagent/committee/pull/145

So I need drop support for old ruby version.

This changes break backward compatibility so we should release 3.0.0.
(3-0 branch for 3.0.0 version)


qsona:
@brandur Hi, I'm +1 on this. We want to merge this PR but I'd like to hear your opinion.

brandur:
@ota42y @qsona @yhirano55 It looks like support for Ruby 2.2 ended ~June 2018 which is roughly 5 months ago. I generally like to keep support around for old versions a little longer than that, but I also see the sense in deprecating things and I don't need <= 2.2 for anything personally, so I don't really have any serious objections to this change.

With regards to OpenAPI 3.0: you guys saw that most of the work has already been done in #128 right? Does #145 supersede that or improve it in some way?

ota42y:
Yes I'm agree to your support opinion.
So when we add change which need drop old version we should do it.
And that is now (when we release OpenAPI3 feature).

ota42y:
Yes @KentaYoshitani is my co-worker and we will create a completely different approach to support OpenAPI3.
So we closed #128

brandur:
> Yes @KentaYoshitani is my co-worker and we will create a completely different approach to support OpenAPI3.
> So we closed #128
>
> Sorry to confuse you

Ah, I see! No worries, and thanks.

ota42y:
We will drop old ruby in 3.0.0

PR #152, opened 2018-11-12
Add Hook triggered when response is invalid

qsona:
## What

When the response is invalid (judged by ResponseValidator), we want to report errors. In this PR, I add a hook point in ResponseValidation Class.

~This PR is work in progress (Although I tested manually on my local PC).~  I added tests and documents.

## Example

```rb
config.middleware.use Committee::Middleware::ResponseValidation, schema: schema,
                        response_error_handler: ->(e) { Raven.capture_exception(e) }
```

## Review Points

- Is there another way to achieve this? (If so, this PR may be needless and should be closed.)
- API
  - Is lambda suitable here?
  - naming of ~`error_hook`~ `response_error_handler`
- Implementation
  - How should we treat if an unexpected error occurred in that lambda

### DIFF
--- a/lib/committee/middleware/response_validation.rb
+++ b/lib/committee/middleware/response_validation.rb
@@ -5,6 +5,7 @@ module Committee::Middleware
     def initialize(app, options = {})
       super
       @validate_errors = options[:validate_errors]
+      @error_handler = options[:error_handler]
     end
 
     def handle(request)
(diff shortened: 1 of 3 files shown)


yhirano55:
@qsona Good idea. How do you feel use `exception_handler` name instead of `error_hook` ?

ota42y:
Looks good!!
But this option looks like all error handling, but your PR handling response error only.
So we need change option name or handle all error.

qsona:
@yhirano55 @ota42y good pointing out! I renamed it to `response_error_handler`.

ota42y:
Please add test :)

qsona:
@ota42y I found that this handler is for only ResponseValidation middleware, so perhaps `response_error_handler` is redundant. How about just `error_handler` as @yhirano55 suggested?

lib/committee/middleware/response_validation.rb
```diff
@@ -5,6 +5,7 @@ class ResponseValidation < Base
     def initialize(app, options = {})
       super
       @validate_errors = options[:validate_errors]
+      @response_error_handler = options[:response_error_handler]
```
yhirano55: @qsona Does it really need a prefix of "response_" ?

PR #155, opened 2018-11-15
Specify newer versions of rack

mathias:
- Continue to use Rack < 2.0 for Rubies < 2.2.2.
- Fixes CVE-2018-16471

Not sure how we feel about being exact with the versions, but this does ensure that the CVE is patched for Rubies newer than 2.2.2.


mathias:
Also, the official CVE doesn't mention 1.6.11, but it looks like from the description of the CVE that it should patch it: 
> There is a possible XSS vulnerability in Rack before 2.0.6 and 1.6.11.

brandur:
Thanks!

brandur:
Released as 2.3.0 along with #147.

PR #163, opened 2018-12-18
replace openapi_parser gem

ota42y:
Use this gem
https://github.com/ota42y/openapi_parser

Because oas_parser gem's parsed result is vagueness.

For example, MediaTypeObject like 1{schema: {type: 'object', properties: {....}}}1.
OasParser::Parameter check root hash 'properties' so we should use schema key
But, array object 'items' properties check ['schema']['items'] so we mustn't flatten :(
Please see this comment https://github.com/interagent/committee/blob/5398d0561701512519759a84513c9f00ffef90dc/lib/committee/schema_validator/open_api_3/operation_object.rb#L67

So I create strict parser.


yhirano55:
@ota42y This code is approved, but it seems CI has not passed. Please check your test codes.

PR #174, opened 2018-12-31
support check_content_type option for OpenAPI3

ota42y:


lib/committee/schema_validator/open_api_3.rb
```diff
@@ -1,5 +1,6 @@
 class Committee::SchemaValidator
-  class OpenAPI3
+    class OpenAPI3
```
yhirano55: [nits] It looks indentation is mistaken.

ota42y: fixed!

PR #201, opened 2019-01-17
Incorrect error and warning messages

amatsuda:
If I understand correctly, `schema:` option for Middleware object accepts an instance of Schema object as the name implies, but the error message and warning message are showing misleading phrases like
"please send a driver object instead" or
"schema expected to be a hash or an instance of Committee::Drivers::Schema".

Here's a quick fix on those messages.


ota42y:
Thank you your contribution!

PR #207, opened 2019-01-22
support validations of Parameter Object

kehra:
The request parameter is insufficiently verified.
So I add some property.

ref: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterObject


brandur:
Thanks @kehra! Looks great.

PR #208, opened 2019-01-22
Bump version to 3.0.0.beta2

brandur:
We ended up changing the naming of the option `filepath` to
`schema_path` after `3.0.0.beta` was released. Here we release a new
version of the 3.0 line that has the new name in it to be consistent
with what's in 2.4.0


brandur:
@ota42y I just released `3.0.0.beta2` on Rubygems. Please take a look at this and merge this into the 3.0 branch when you get a chance. Thanks!

PR #209, opened 2019-01-23
Fix pattern value

kehra:
Fix #207

I'm sorry.
Pattern value should be Regexp object.
ref: https://github.com/brandur/json_schema/blob/f07e6a8f9a608c132b6b2bb05793607e68005c9e/lib/json_schema/parser.rb#L244-L253


brandur:
Thanks @kehra / @ota42y!

Released as 2.5.1.

kehra:
thx! @brandur @ota42y 🙇

PR #210, opened 2019-01-24
OpenAPI 3 merge request body to request parameter in GET request

ota42y:
When the user send parameter in requset body, committee with Hyper-Schema use it as requset parameter.
Because RequestUnpacker is parsed request body on all HTTP method.
https://github.com/interagent/committee/blob/master/lib/committee/request_unpacker.rb#L17

But when we use OpenAPI 3, we check request parameter only so we get error when required paramater in request body.
[OpenAPI 3 definition are not allowed request body](https://swagger.io/docs/specification/describing-request-body/) and request body isn't request parameter so it's correct behaviour.

```
GET, DELETE and HEAD are no longer allowed to have request body 
because it does not have defined semantics as per RFC 7231.
```

But in the committee, I think we should merge to request parameter for validation.
(This means OpenAPI 3 check requset parameter and request body like Hyper-Schema)

Because if we changee this behaviour, we should fix all client application when upgroade Hyper-Schema or OpenAPI 2 to OpenAPI3.
Sometimes the client application isn't under their control for example when APIs are published to internet so it's change isn't good.

In [RFC 7231](https://tools.ietf.org/html/rfc7231#section-4.3), there is no defined semantics.
So I think this behavior is bad but not very wrong.
```
   A payload within a GET request message has no defined semantics;
   sending a payload body on a GET request might cause some existing
   implementations to reject the request.
```

Now OpenAPI 3 validate and coerce merged parameter and bodies.
But when we coerce value, we check again with parameters only so we get error if the parameter in request body.
So I fix it to copy coerced value from merged data.

### DIFF
--- a/lib/committee/schema_validator/open_api_3.rb
+++ b/lib/committee/schema_validator/open_api_3.rb
@@ -17,7 +17,7 @@ class Committee::SchemaValidator
 
       request_schema_validation(request)
 
-      @operation_object&.coerce_request_parameter(request.env["rack.request.query_hash"], header(request), validator_option) if !request.GET.nil? && !request.env["rack.request.query_hash"].empty?
+      copy_coered_data_to_query_hash(request)
     end
 
     def response_validate(status, headers, response, test_method = false)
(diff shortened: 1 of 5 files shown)


lib/committee/schema_validator/open_api_3.rb
```diff
@@ -17,7 +17,7 @@ def request_validate(request)
 
       request_schema_validation(request)
 
-      copy_coered_data_to_query_hash(request)
+      copy_coerced_data_to_query_hash(request)
```
yhirano55: 👍

PR #212, opened 2019-01-25
3.0.0.beta3

ota42y:


ota42y:
@brandur I released 3.0.0.beta3 on rubygems.org, because this change include corner case fix only :)

PR #216, opened 2019-03-18
Correct `filepath` => `schema_path`

qsona:
The `filepath` option has been replaced to `schema_path`, but I faced to an old error message so I've fixed it. I also made same kind of fix in test files.


brandur:
Thanks @qsona!

brandur:
Released as 3.0.1.

qsona:
@brandur Thank you for merging it. I also appreciate your work as the owner even though you're not using this library now.

brandur:
No worries at all @qsona! (And thanks! :)

PR #225, opened 2019-06-17
Catch OpenAPIParser::NotExistRequiredKey with RequestValidation

chibicco:
日本語で申し訳ありませんが、ご確認よろしくお願い致します🙇‍♂️

## Context

`Committee::ValidationError` を継承したクラスで、`OpenAPIParser::NotExistRequiredKey` 発生時にエラーをキャッチできない。
(request parameter に required key が不足している場合に発生する)

### DIFF
--- a/lib/committee/schema_validator/open_api_3/operation_wrapper.rb
+++ b/lib/committee/schema_validator/open_api_3/operation_wrapper.rb
@@ -18,6 +18,8 @@ module Committee
       return {} unless options.coerce_value
 
       request_operation.validate_path_params(options)
+    rescue OpenAPIParser::NotExistRequiredKey => e
+      raise Committee::InvalidRequest.new(e.message)
     end
 
     # @param [Boolean] strict when not content_type or status code definition, raise error
--- a/test/middleware/request_validation_open_api_3_test.rb
+++ b/test/middleware/request_validation_open_api_3_test.rb
@@ -312,6 +312,16 @@ describe Committee::Middleware::RequestValidation do
     assert_equal 404, last_response.status
   end
 
+  it "OpenAPI3 parser not exist required key" do
+    @app = new_rack_app(raise: true, schema: open_api_3_schema)
+
+    e = assert_raises(Committee::InvalidRequest) do
+      get "/validate", nil
+    end
+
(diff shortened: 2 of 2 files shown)


lib/committee/middleware/request_validation.rb
```diff
@@ -16,7 +16,7 @@ def handle(request)
       raise Committee::NotFound, "That request method and path combination isn't defined." if !schema_validator.link_exist? && @strict
 
       @app.call(request.env)
-    rescue Committee::BadRequest, Committee::InvalidRequest
+    rescue Committee::BadRequest, Committee::InvalidRequest, OpenAPIParser::NotExistRequiredKey
```
ota42y: Sorry, this class is a common part so we don't uses OpenAPI or JSON Hyper-Schema specific code.
So we should rescue in Committee::SchemaValidator::OpenAPI3::OperationWrapper class and we need to transfer the error message.

ota42y: When the committee call OpenAPIParser method, rescue error and transfer the error message to committee's error class.
https://github.com/interagent/committee/blob/master/lib/committee/schema_validator/open_api_3/operation_wrapper.rb#L98-L103
https://github.com/interagent/committee/blob/master/lib/committee/schema_validator/open_api_3/operation_wrapper.rb#L110-L112

But we doesn't rescue in path parameter validation.
I think we can solve this problem to rescue error and transfer Committee::InvalidRequest class.
https://github.com/interagent/committee/blob/master/lib/committee/schema_validator/open_api_3/operation_wrapper.rb#L20

chibicco: Tahnks!

I fixed it, And rebased.

chibicco: Thanks!

I fixed it, And rebased.
https://github.com/interagent/committee/pull/225/commits/4308115d6e26156d82720915384b64df422791bc

ota42y:
It's ok 🙆 
Thank you great fix !!!!!

ota42y:
I just released 3.0.3! Thanks!!!
https://rubygems.org/gems/committee/versions/3.0.3

PR #227, opened 2019-07-16
Support request validations by assert_schema_conform

chibicco:
## Issue

With assert_schema_conform, you can only confirm the response.

## This PR

I can't find request check, so I added it.

## Example

Can performa request validation.

```
assert_schema_conform(request_validation: true, response_validation: false)
```

### DIFF
--- a/lib/committee/test/methods.rb
+++ b/lib/committee/test/methods.rb
@@ -1,18 +1,22 @@
 module Committee::Test
   module Methods
     def assert_schema_conform
-      @schema ||= Committee::Middleware::Base.get_schema(committee_options)
-      @router ||= @schema.build_router(committee_options)
+      assert_request_schema_confirm unless old_behavior
+      assert_response_schema_confirm
+    end
 
-      v = @router.build_schema_validator(request_object)
+    def assert_request_schema_confirm
+      schema_validator.request_validate(request_object)
+    end
 
-      unless v.link_exist?
+    def assert_response_schema_confirm
+      unless schema_validator.link_exist?
         response = "`#{request_object.request_method} #{request_object.path_info}` undefined in schema."
         raise Committee::InvalidResponse.new(response)
       end
 
       status, headers, body = response_data
-      v.response_validate(status, headers, [body], true) if validate_response?(status)
+      schema_validator.response_validate(status, headers, [body], true) if validate_response?(status)
     end
 
     def committee_options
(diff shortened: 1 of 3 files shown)


ota42y:
It's very good changes!!!

But I think it's good to define two new method than options.
Because I want to avoid that the test method changes its behavior by options

```
assert_response_schema_confirm
assert_request_schema_confirm
```

And I think we offen want to assert same time.
So the `assert_schema_confirm` method should check request and response.
However, we should keep backward compatible

Now assert_schema_confirm and assert_response_schema_confirm is same method.
So we should show warning message and can keep existing behavior with minimal changes (like setting options)
And when we release new version, we delete that's options and assert request and response 

like this...

```
def assert_schema_confirm
  old_behavior = options[:old_assert_behavior]
  if old_behavior.nil?
    # when the user not set option, show message like..
    # now assert_schema_confirm check response schema only.
    # but we will change check request and response in futer mejor version.
    # so if you want to conform response only, please use assert_response_schema_confirm
    # or you can supress this message and keep old behavior by setting old_assert_behavior=true 
    old_behavior = true
  end

  assert_request_schema_confirm unless old_behavior
  assert_response_schema_confirm
end
```

chibicco:
I just added tests for each method but didn't add it for options.
That is just because that it seems to me that adding tests covered for options is too much this change.
Besides, I did force-push because of making commits arranged. 

Thank you.

lib/committee/test/methods.rb
```diff
@@ -1,18 +1,22 @@
...
     def assert_schema_conform
-      @schema ||= Committee::Middleware::Base.get_schema(committee_options)
-      @router ||= @schema.build_router(committee_options)
+      assert_request_schema_confirm unless old_behavior
+      assert_response_schema_confirm
+    end
 
-      v = @router.build_schema_validator(request_object)
+    def assert_request_schema_confirm
+      schema_validator.request_validate(request_object)
```
ota42y: I think we should display an error if this endpoint's schema is not defined like line 13.
(Please be careful it's different in the endpoint which does not have request schema definition and the endpoint's schema doesn't exist)

chibicco: It appeared to be returning in the request_schema_validation method.
Could you tell me a little more if you should do another check?

https://github.com/interagent/committee/blob/master/lib/committee/schema_validator/open_api_3.rb#L54
https://github.com/interagent/committee/blob/master/lib/committee/schema_validator/hyper_schema.rb#L78

chibicco: Thank you for the polite explanation!
I tried it, And added a test of request and response.
https://github.com/interagent/committee/pull/227/commits/7316598680c5fd808f1547dc112342262a5d7cd3

PR #228, opened 2019-07-18
fix: ensure we catch all OpenAPIParser::OpenAPIError classes when coercing path parameters for OpenAPI 3

mt-kelvintaywl:
#### bug summary

I found that when we send a request with path
```
/resource/{id}/...
```
 
while a missing ID will be wrapped into a `Committee::InvalidRequest error`, an existing but invalid ID will still be uncaught; This gets bubbled up as an `OpenAPIParser::ValidateError` instead.

I think for consistency, we can catch all `OpenAPIParser::OpenAPIError` in the rescue operation.

Related PR: https://github.com/interagent/committee/pull/225


ota42y:
Oh our error handling was not good, thank you!

PR #230, opened 2019-08-07
Add error handler option to Committee::Middleware::RequestValidation

yhirano55:
Added error_handler option to `Committee::Middleware::RequestValidation`

### Use case

When request is invalid, I want to send an exception event to Sentry. But currently it cound not do that. So I've opened this pull request.


lib/committee/middleware/response_validation.rb
```diff
@@ -5,8 +5,6 @@ class ResponseValidation < Base
     def initialize(app, options = {})
       super
       @validate_success_only = @schema.validator_option.validate_success_only
-
-      @error_handler = options[:error_handler]
```
ota42y: 💯

PR #233, opened 2019-08-10
Refactor code layout

yhirano55:
## Summary

- It looks it's better to load children after parent module or class.
- Currently it's a bit of hard to read code. So I've refactored them.


lib/committee.rb
```diff
@@ -55,3 +21,13 @@ def self.warn_deprecated(message)
...
+
+require_relative "committee/drivers"
+require_relative "committee/errors"
+require_relative "committee/middleware"
+require_relative "committee/request_unpacker"
+require_relative "committee/schema_validator"
+require_relative "committee/validation_error"
+
+require_relative "committee/bin/committee_stub"
+require_relative "committee/test/methods"
```
ota42y: very good!!! 💯

PR #240, opened 2019-09-04
OS3 request validator skip if link does not exist

yhirano55:
## Summary

- When link does not exist in schema.json, it seems that request object coerced change with unpack.
- It looks it's not intentional behavior, so I've added guard clause  which checks link exists or not before validation.

### DIFF
--- a/lib/committee/schema_validator/open_api_3.rb
+++ b/lib/committee/schema_validator/open_api_3.rb
@@ -12,6 +12,8 @@ module Committee
       end
 
       def request_validate(request)
+        return unless link_exist?
+
         path_params = validator_option.coerce_path_params ? coerce_path_params : {}
 
         request_unpack(request)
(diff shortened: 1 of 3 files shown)


test/middleware/request_validation_open_api_3_test.rb
```diff
@@ -256,7 +256,7 @@ def app
   it "rescues JSON errors" do
     @app = new_rack_app(schema: open_api_3_schema)
     header "Content-Type", "application/json"
-    post "/apps", "{x:y}"
+    post "/characters", "{x:y}"
```
yhirano55: This path doesn't exist, it returns 200. To test JSON errors, I changed existing path. :)

lib/committee/schema_validator/open_api_3.rb
```diff
@@ -49,7 +51,6 @@ def coerce_form_params(_parameter)
       attr_reader :validator_option
 
       def coerce_path_params
-        return {} unless link_exist?
```
ota42y: 💯

test/middleware/request_validation_open_api_3_test.rb
ota42y: 👍

PR #244, opened 2019-10-10
Bugfix: validate request parameter for POST, PUT, PATCH operations

maurobellati:
Currently request parameters are validated just for GET and DELETE.

Add request parameters validation even for POST, PUT, PATCH.

### DIFF
--- a/test/data/openapi3/normal.yaml
+++ b/test/data/openapi3/normal.yaml
@@ -599,7 +599,7 @@ components:
   responses:
     header_integer_required:
       headers:
-        integer:
+        'x-limit':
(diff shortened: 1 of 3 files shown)


test/data/openapi3/normal.yaml
```diff
@@ -502,18 +502,35 @@ paths:
   /header:
     get:
       parameters:
-        - name: integer
-          in: header
-          required: true
-          schema:
-            type: integer
+        - $ref: '#/components/parameters/header_integer_required'
```
ota42y: 💯

PR #256, opened 2020-03-18
Allow path param and request body to be passed at the same time

jalyna:
Right now it is not possible in open api 3 to set a path param and a request body.

Not sure though if this is the right way but it worked for the spec we are using.

### DIFF
--- a/lib/committee/request_unpacker.rb
+++ b/lib/committee/request_unpacker.rb
@@ -14,6 +14,48 @@ module Committee
     end
 
     def call
+      return call_hyperschema if hyperschema?
+
+      # if Content-Type is empty or JSON, and there was a request body, try to
+      # interpret it as JSON
+      params = {}
+
+      params['body'] = if !@request.media_type || @request.media_type =~ %r{application/.*json}
+        parse_json
+      elsif @optimistic_json
+        begin
+          parse_json
+        rescue JSON::ParserError
+          nil
+        end
+      else
+        {}
+      end
+
+      params['form_data'] = if @allow_form_params && %w[application/x-www-form-urlencoded multipart/form-data].include?(@request.media_type)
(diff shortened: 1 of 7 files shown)


kurko:
@ota42y can you be more specific? Why would that break older versions?

I have an interest in this PR and would like to help move it forward.

ota42y:
The committee support convert datetime string to datetime object.

If we don't use committee with this code (Ruby on Rails),  
we'll get ArgumentError because `params[:registered_at]` is string and Time.now is Time class.
But the committee convert `params[:registered_at]` to DateTime, so this code is valid.


```ruby
class UsersController < ApplicationController
    skip_before_action :verify_authenticity_token

    def create
        binding.pry
        registered_at = params[:registered_at]
        raise "error" unless registered_at < Time.now
        user = ["debug user1"] # do something

        head :ok
    end
end
```

When we use this feature on rails, we need this config to set converted value to `ActionController::Parameters`.
```
    schema_path = Rails.root.join("config/openapi.yaml").to_s
    config.middleware.use Committee::Middleware::RequestValidation, 
      schema_path: schema_path,
      params_key: 'action_dispatch.request.request_parameters'
```

But this change break this code because this code pass nested hast to `ActionController::Parameters`.
- old version
  - {"registerd_at": DateTime Object}
- this version 
  - {"body": {"registered_at": DateTime Object}}

https://gist.github.com/ota42y/0870100c03444d39d78107a23de9f171

hedselu:
Any update on this issue?

makdad:
@ota42y I believe with 5.0 released, this issue is "fixable" in the sense that backward compatibility is less of an issue for the hash keys.  Could you confirm?  This is a pretty big issue for Open API Spec compatibility IMO.

geemus:
Open to discussion if others would like to pursue this further.

PR #258, opened 2020-03-24
fix: parse response body as JSON only when applicable

### LINKED ISSUE #257: Support for non-JSON response validation
I have an OpenAPI 3 spec that for one of the responses return a HTTP 302 response, for example:

```yaml
# snippet
paths:
    /faq/{id}:
    get:
      description: |
        fetch a FAQ page (HTTP 302 redirect to our external CMS for FAQ page)
      responses:
        '302':
          description: Response of a Connect redirect
          headers:
            Location:
              schema:
                type: string
               description: Redirection URL
           content:
             text/plain:
               schema:
                 type: string
```

and since OpenAPI3's response_validate handler will attempt to always parse response body to JSON:
https://github.com/interagent/committee/blob/fe8570af3c54cb467b857350f1aeec8444c3f9a3/lib/committee/schema_validator/open_api_3.rb#L28-L39

This is the same for HyperSchema too:
https://github.com/interagent/committee/blob/fe8570af3c54cb467b857350f1aeec8444c3f9a3/lib/committee/schema_validator/hyper_schema.rb#L31-L40

I ended up having to set `validate_success_only` to true, but that meant I turned off validation for all non HTTP 2XX responses including client errors (HTTP 4XX) :(

Would be nice for the response_validate to attempt to **only** parse as JSON based on the Content-Type response header?

mt-kelvintaywl:
This resolves #257

I had a use-case where I am using the Committee::Middleware::ResponseValidation middleware with `strict: true` option, and one of my endpoint returns a HTTP 302 response (with Location header) and has a text/plain content as string ('You are being redirected to ....' as per Rails default).

My OpenAPI 3 spec reflects this correctly. However, it seems the ResponseValidation always assumes the response body is parseable as JSON (not always the case).

I added the code I think would be sufficient to call on `JSON.parse` only when the headers suggest to.
For backwards compatability, this behaviour is set only when `parse_response_by_content_type` is set to `true` (defaults `false`).

I think perhaps this can be improved, and I would like to add some tests as least 🙇 
As of now, `bundle exec rake` command exits 0 on my local testing (ruby 2.6)

### DIFF
--- a/lib/committee/schema_validator/option.rb
+++ b/lib/committee/schema_validator/option.rb
@@ -15,7 +15,8 @@ module Committee
                   :coerce_query_params,
                   :coerce_recursive,
                   :optimistic_json,
-                  :validate_success_only
+                  :validate_success_only,
+                  :parse_response_by_content_type
 
       # Non-boolean options:
       attr_reader :headers_key,
@@ -35,6 +36,7 @@ module Committee
         @check_header        = options.fetch(:check_header, true)
         @coerce_recursive    = options.fetch(:coerce_recursive, true)
         @optimistic_json     = options.fetch(:optimistic_json, false)
+        @parse_response_by_content_type = options.fetch(:parse_response_by_content_type, false)
 
         # Boolean options and have a different value by default
         @allow_get_body      = options.fetch(:allow_get_body, schema.driver.default_allow_get_body)
--- a/README.md
+++ b/README.md
@@ -153,6 +153,7 @@ Option values and defaults:
 |raise| false | false | Raise an exception on error instead of responding with a generic error body. |
 |validate_success_only| true | false | Also validate non-2xx responses only. |
 |ignore_error| false | false | Validate and ignore result even if validation is error. So always return original data. |
+|parse_response_by_content_type| false | false | Parse response body to JSON only if Content-Type header is 'application/json'. When false, this always optimisitically parses as JSON without checking for Content-Type header. |
 
 No boolean option values:
 
--- a/lib/committee/schema_validator/hyper_schema.rb
+++ b/lib/committee/schema_validator/hyper_schema.rb
@@ -35,7 +35,12 @@ module Committee
         response.each do |chunk|
           full_body << chunk
         end
-        data = full_body.empty? ? {} : JSON.parse(full_body)
+
+        data = {}
+        if !full_body.empty? && headers.fetch('Content-Type', nil)&.start_with?('application/json')
+          data = JSON.parse(full_body)
+        end
+
         Committee::SchemaValidator::HyperSchema::ResponseValidator.new(link, validate_success_only: validator_option.validate_success_only).call(status, headers, data)
       end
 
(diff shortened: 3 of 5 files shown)


mt-kelvintaywl:
@ota42y 

Thank you for the response and feedback!

Indeed, i agree with you and added a config option for ResponseValidation `parse_response_by_content_type`.

will add new tests to ensure behaviour!

lib/committee/schema_validator/option.rb
```diff
@@ -35,6 +36,7 @@ def initialize(options, schema, schema_type)
         @check_header        = options.fetch(:check_header, true)
         @coerce_recursive    = options.fetch(:coerce_recursive, true)
         @optimistic_json     = options.fetch(:optimistic_json, false)
+        @parse_response_by_content_type = options.fetch(:parse_response_by_content_type, false)
```
ota42y, line 39: I think this option should be true but we can't set for backward compatibility.
So when we upgrade major version, we should change default value.

But the user doesn't set option will get error.
So when the user doesn't set value, we should display a warning message like `please set parse_response_by_content_type = false because we'll change default value in next major version` to make the upgrade easier.

Please use `Committee.warn_deprecated` to display warning message
https://github.com/interagent/committee/blob/10734081293b7df808ded7e9a06a9bf97198a67b/lib/committee/middleware/base.rb#L79

ota42y: - parse_response_by_content_type = false
  - old behavior
- parse_response_by_content_type  = true
  - new behavior
- parse_response_by_content_type  = nil
  - show warning message and old behavior

fruwe:
Maybe not directly related, but there is also the case when you want to validate a file.

E.g. the schema could be
```
responses: 
  200:
    description: ok
    content:
      text/csv:
        schema:
          type: string
          format: binary
```

and in the controller you could have something like this:

```
  send_data(
    csv_string,
    type: "text/csv; charset=utf-8; header=present",
    disposition: "attachment; filename=#{file_name}.csv",
  )
```

which currently throws an JSON::ParserError or Committee::InvalidResponse using this PR

mt-kelvintaywl:
ah, apologiess for the late response.
unfortunately, i've moved on to a new project that does not use `committee` so I have not had time to come back to this  🙇 
would be happy if someone else would like to pick this up! :D

kinoppyd:
Is this PR aborted?
I'm really need this behavior, so what can I do to help merge this PR?

adamaiken89:
Hi how is the current progress?
I think this change is necessary as there are a lot of cases for not using json.

If you need maintainer I can take this.

ota42y:
mt-kelvintaywl said `would be happy if someone else would like to pick this up! :D`
So I think anyone can use these commits to continue.

So if adamaiken89 's going to take over, I think it's better to continue with that PR. :) (thanks !)

adamaiken89:
I have created another PR based on this one. Please review and comment in that PR.

ota42y:
Thank you!
LGTM

I'm going to merge this change in tonight !

ota42y:
This change was merged in https://github.com/interagent/committee/pull/273

PR #273, opened 2020-06-23
Fix json parse only if applicable

adamaiken89:
This is the pull request based on https://github.com/interagent/committee/pull/258 . 
I made some modification so we can pass body to `openapi_validator` even if it is not a json object.

### DIFF
--- a/lib/committee/schema_validator/open_api_3.rb
+++ b/lib/committee/schema_validator/open_api_3.rb
@@ -30,7 +30,11 @@ module Committee
         response.each do |chunk|
           full_body << chunk
         end
-        data = full_body.empty? ? {} : JSON.parse(full_body)
+
+        data = {}
+        if !full_body.empty? && headers.fetch('Content-Type', nil)&.start_with?('application/json')
+          data = JSON.parse(full_body)
+        end
 
         strict = test_method
         Committee::SchemaValidator::OpenAPI3::ResponseValidator.
--- a/lib/committee/schema_validator/open_api_3.rb
+++ b/lib/committee/schema_validator/open_api_3.rb
@@ -32,8 +32,9 @@ module Committee
         end
 
         data = {}
-        if !full_body.empty? && headers.fetch('Content-Type', nil)&.start_with?('application/json')
-          data = JSON.parse(full_body)
+        unless full_body.empty?
+          parse_to_json = !validator_option.parse_response_by_content_type || headers.fetch('Content-Type', nil)&.start_with?('application/json')
+          data = JSON.parse(full_body) if parse_to_json
         end
 
         strict = test_method
--- a/lib/committee/schema_validator/open_api_3.rb
+++ b/lib/committee/schema_validator/open_api_3.rb
@@ -31,10 +31,12 @@ module Committee
           full_body << chunk
         end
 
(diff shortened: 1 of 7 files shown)


lib/committee/schema_validator/open_api_3.rb
```diff
@@ -31,10 +31,12 @@ def response_validate(status, headers, response, test_method = false)
...
-        unless full_body.empty?
-          parse_to_json = !validator_option.parse_response_by_content_type || headers.fetch('Content-Type', nil)&.start_with?('application/json')
-          data = JSON.parse(full_body) if parse_to_json
+        parse_to_json = !validator_option.parse_response_by_content_type || 
+                        headers.fetch('Content-Type', nil)&.start_with?('application/json')
+        data = if parse_to_json
+          full_body.empty? ? {} : JSON.parse(full_body)
+        else
+          full_body
         end
```
mt-kelvintaywl: just thinking if this is readable enough and succinct
```suggestion
        jsonable = !validator_option.parse_response_by_content_type || 
                        headers.fetch('Content-Type', nil)&.start_with?('application/json')
        data = jsonable ? (full_body.empty? ?  {} : JSON.parse(full_body)) : full_body
```

lib/committee/schema_validator/option.rb
```diff
@@ -36,7 +36,12 @@ def initialize(options, schema, schema_type)
         @check_header        = options.fetch(:check_header, true)
         @coerce_recursive    = options.fetch(:coerce_recursive, true)
         @optimistic_json     = options.fetch(:optimistic_json, false)
-        @parse_response_by_content_type = options.fetch(:parse_response_by_content_type, false)
+        @parse_response_by_content_type = if options[:parse_response_by_content_type].nil?
+          Committee.warn_deprecated('Committee: please set parse_response_by_content_type = false because we\'ll change default value in next major version.') 
+          false
+        else
+          options.fetch(:parse_response_by_content_type)
+        end
```
mt-kelvintaywl, line 44: hmm, maybe @ota42y has a better answer, but since we do not practice branching per major versions, and that we have already released version 4.0.0 (major version), i wonder if we can simply have `parse_response_by_content_type` use `false` by default?
> in other words, the new `parse_response_by_content_type` is only available for >= 4.0.0 onwards? 

in any case, i think the message :
> because we'll change default value in next major version.
can be improved by saying what major version that would be.

lib/committee/schema_validator/open_api_3.rb
adamaiken89: It looks a little bit confusing to use two ternary operators in the same line though...

lib/committee/schema_validator/option.rb
ota42y: I think we should show error message.

This change is better behavior, beneficial for many new users so we should set `true` by default
But we can't break backward compatibility for now users....

So I think we set `false` by default in this version and show warning message the user doesn't set value.
If the user responds to all warning messages, changing the default value will not affect the user, so no one is in trouble.

ota42y:
(I wrote a reply but it seems that I could not send it.)

LGTM! Thanks to great changes!!!!!!!!!!!

PR #278, opened 2020-08-21
Don't rescue JSON::ParserError

### LINKED ISSUE #277: Though request format is HTML (not JSON), ResponseValidation handles JSON::ParserError and renders json
## Expected

- Skip processing if request format is not JSON

## Actual

1. Raise `JSON::ParserError` in `@app.call(request.env)`
2.  Handle its exception
3. Then this middleware render an error response with json format.

https://github.com/interagent/committee/blob/3a0ed159c8c798c15f75d2e71057d0f86d46875f/lib/committee/middleware/response_validation.rb#L3-L33

yhirano55:
Resolve #277

When app server provides multiple format not only json but html or xml,
it's inconvenient that response validation middleware handles `JSON::ParserError`.

So I thoght response response middleware should not handle `JSON::ParserError` exception.

@ota42y What do you think about it ?

### DIFF
--- a/lib/committee/middleware/response_validation.rb
+++ b/lib/committee/middleware/response_validation.rb
@@ -22,11 +22,6 @@ module Committee
 
           raise if @raise
           return @error_class.new(500, :invalid_response, $!.message).render unless @ignore_error
-        rescue JSON::ParserError
-          handle_exception($!, request.env)
-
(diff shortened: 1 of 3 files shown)


yhirano55:
@ota42y You're right. I'll expect this bug is fixed soon by your work. So, I close this PR, thanks.

PR #281, opened 2020-09-02
Hold original exception only openapi3

uplus:
I want to handle the OpenAPI3 original exceptions for more advanced error handling.


ota42y:
It's a very good PR! !!
The other errors are created by the committee, so we need add only that error classes.

Thank you very much!

PR #287, opened 2020-09-28
Make check for complex JSON types more strict

knpwrs:
#64 introduced the ability to parse complex JSON types such as `application/vnd.api+json`. Unfortunately, the check implemented breaks `committee` for certain extensions of JSON such as [`application/x-ndjson`](http://ndjson.org/).

This PR attempts to make the check more strictly match the format mentioned in #64 by ensuring that anything preceding `json` ends with `+`. An alternative suggestion would be to not unpack requests with a content type of `application/x-.+`.


ota42y:
Thanks!!!!!!!! 👍 👍 👍 👍 👍 👍 👍 👍

PR #291, opened 2020-11-08
Improve deprecation messages

r7kamura:
## Background

It's easier to understand if all the warning messages from committee are written in the same format like `[DEPRECATION] ...`, and I think this logic should be implemented using the same method.

I found `Committee.warn_deprecated` method for this purpose, so I tried to replace all deprecation warnings logic with this method.

### Committee::SchemaValidator::Option#initialize

This will affect the following message on this method:

> Committee: please set parse_response_by_content_type = false because we\'ll change default value in next major version.

chaning it to:

> [DEPRECATION] Committee: please set parse_response_by_content_type = false because we\'ll change default value in next major version.

### Committee::Test::Methods#old_behavior

Added a minor capitalization change on this method, chaning:

> [DEPRECATION] now assert_schema_conform check...

to:

> [DEPRECATION] Now assert_schema_conform check...

### DIFF
--- a/lib/committee.rb
+++ b/lib/committee.rb
@@ -16,9 +16,7 @@ module Committee
   end
 
   def self.warn_deprecated(message)
-    if !$VERBOSE.nil?
-      $stderr.puts(message)
-    end
+    warn(message)
   end
 end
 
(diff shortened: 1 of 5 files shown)


ota42y:
Nice changes!!!!! 👍 👍 👍 👍

PR #295, opened 2020-12-14
add request to validation error

kentakag:
Currently, when I try to customize the error_body of the ValidationError class, it can only contain id, status and message.
I would like to include the request object in this error_body, because I especially want to include the request_id in the error_body.


ota42y:
Thank you for your feedback !!!!!!

PR #301, opened 2021-02-12
Disable coerece_form_params option in the case of application/json

### LINKED ISSUE #286: problem with oneOf: string/integer
I have added a rails application here that illustrates the problem: https://github.com/jcoyne/comittee-bug

It has a failing test which you can run by doing:
```
bundle exec rspec spec/requests/resources_request_spec.rb
```

Committee reports`{"id":"bad_request","message":"7116 isn't one of in #/components/schemas/Title/properties/value"}`

I believe this is because of: https://github.com/jcoyne/comittee-bug/blob/master/openapi.yml#L32-L34

If I comment out the line:
```
          - type: integer
```

then it works fine.

If I change this line of the test:

https://github.com/jcoyne/comittee-bug/blob/master/spec/requests/resources_request_spec.rb#L6

to look like a string, e.g. (`{ "title":{"value":"Number 7116"}}`) it also works fine.

meganemura:
To fix #286.

This disables coerece_form_params option in the case of `application/json`.
But, disabling coerce_form_params prevents coercing DateTime in object.

https://github.com/ota42y/openapi_parser/blob/d17334a866b39fa197d324b9d321df2b03d469b6/lib/openapi_parser/schema_validators/object_validator.rb#L41

I found that defining `ObjectValidator#initialize(validator, coerce_value, datetime_coerce_class)` in OpenAPIParser and adding coercing condition like below, it works. but I'm not sure this way is good or not.

```diff
-value.merge!(coerced_values.to_h) if @coerce_value
+value.merge!(coerced_values.to_h) if @coerce_value || !@datetime_coerce_class.nil?
```

### DIFF
--- a/lib/committee/schema_validator/open_api_3/operation_wrapper.rb
+++ b/lib/committee/schema_validator/open_api_3/operation_wrapper.rb
@@ -90,8 +90,12 @@ module Committee
         end
 
         # @return [OpenAPIParser::SchemaValidator::Options]
-        def build_openapi_parser_post_option(validator_option)
-          coerce_value = validator_option.coerce_form_params
+        def build_openapi_parser_post_option(validator_option, content_type)
+          coerce_value = if content_type =~ %r{application/(?:.*\+)?json}
+                           false
+                         else
+                           validator_option.coerce_form_params
+                         end
           datetime_coerce_class = validator_option.coerce_date_times ? DateTime : nil
           validate_header = validator_option.check_header
           OpenAPIParser::SchemaValidator::Options.new(coerce_value: coerce_value,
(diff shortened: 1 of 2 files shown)


ota42y:
Thanks for the PR.
I think this PR is good, because we should not converting json by `coerce_form_params=true`.

When we use form/path paramemter integer will be convert string so we need string to integer.
However we don't need convet JSON because it have a types so I'd like to make this behavior the default in next major version (we shouldn't break backward compatibility).
So we need create an option called `corece_json_params`, and if it is not set, we should show a warning message and set 'true' (we will change default `false` in next version)

But, converting the datetime format string to Datetime object is useful for JSON users, so I think many user will want to use it.
So we should enable this feature separately from the `corece_json_params` disable.

To summarize
- Do not coerce json by `coerce_form_param=true`
  - This PR is goodl :+1:
- Add `coerce_json_param`
  - We can keep the current behavior by default
  - We should change default `false` in next version
- Rename coerce_date_times to convert_stirng_by_datetime_formant
  - do a convert regardless of coerce_json_param=false (requires a change to openapi_parser)
  - Use `coerce` to convert string to OpenAPI definition types, and `convert` to change from OpenAPI definition types to Ruby's object.

PR #302, opened 2021-03-29
Test Assertions: assert HTTP status code explicitly

skryukov:
This PR adds explicit assertion of HTTP status code instead of use of `validate_success_only` attribute.

before:

```ruby
assert_schema_conform
assert_response_schema_confirm
```

after:
```ruby
assert_schema_conform(200)
assert_response_schema_confirm(200)
```

### Context

The current implementation does not provide a way to assure that the request's preparation was successful and the response is checked against the correct scheme. The `validate_success_only` parameter partially solves this problem, but only for successful requests. However, if we want to check the error code, we will again find ourselves in a possible error situation. Explicit assertion of HTTP status code solves this issue.

### DIFF
--- a/lib/committee/test/methods.rb
+++ b/lib/committee/test/methods.rb
@@ -3,9 +3,9 @@
 module Committee
   module Test
     module Methods
-      def assert_schema_conform
+      def assert_schema_conform(expected_status = nil)
         assert_request_schema_confirm unless old_behavior
-        assert_response_schema_confirm
+        assert_response_schema_confirm(expected_status)
       end
 
       def assert_request_schema_confirm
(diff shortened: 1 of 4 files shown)


lib/committee/test/methods.rb
```diff
@@ -17,14 +17,21 @@ def assert_request_schema_confirm
         schema_validator.request_validate(request_object)
       end
 
-      def assert_response_schema_confirm
+      def assert_response_schema_confirm(expected_status = nil)
```
ota42y, line 20: I thought it would be better to set a default value, but since it is a feature that does not exist originally, there is no point in maintaining compatibility, and this method of setting by the user is good.

lib/committee/test/methods.rb
```diff
@@ -17,14 +17,21 @@ def assert_request_schema_confirm
...
         unless schema_validator.link_exist?
           response = "`#{request_object.request_method} #{request_object.path_info}` undefined in schema (prefix: #{committee_options[:prefix].inspect})."
           raise Committee::InvalidResponse.new(response)
         end
 
         status, headers, body = response_data
 
+        if expected_status.nil?
+          Committee.warn_deprecated('Pass expected response status code to check it against the corresponding schema explicitly.')
```
ota42y, line 29: Nice!!!

ota42y (approved):

PR #315, opened 2021-05-03
support types (alpha version)

ota42y:
add types.
(but auto generate by rbs so this is alpha version)


ota42y:
We need drop 2.4, 2.5 so we should update major version 🤔

ota42y:
This change is very old so close it and retry implement

Issue #317, opened 2021-05-10
Raise error in strict mode when missing response code


hiddewie:
This is a feature request.

Let's assume we have an OpenAPI specification with an API which is tested well. I run Committee in strict mode, such that Committee raises errors in case of missing path/method combinations. This works perfectly.

However, responses with a HTTP response (status) code that is not documented are ignored. In addition, no content validation is performed which can introduce breaking changes in the API that go unnoticed.

Instead, when a response is validated by Committee for which no response code is documented, an error should be raised.

Thanks in advance.

ota42y:
Thank you your feedback!!!

Currently response validation does not support the strict option.
https://github.com/interagent/committee#committeemiddlewareresponsevalidation

I think some users set strict option to response validation and these users will get error when we support strict option.
So we should major update for supporting this option.
(We are planning 5.0.0 release so this change will include this version)

hiddewie:
Thanks for the reaction. 

If people set configuration that is unsupported (`strict` on response validation), it is their probem right? I imagine it is impossible to be backwards compatible for undefined configuration.

Maybe they even expect strict response code validation when they set `strict` to true, even when it is not implemented right now.

PR #319, opened 2021-05-15
strict response validation

ota42y:
This changes break backward compatibility because when the user set strict option, old version is ignore but this version doesn't ignore.
So we should add warning on old version.

close: https://github.com/interagent/committee/issues/317

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,7 @@ Option values and defaults:
 |validate_success_only| true | false | Also validate non-2xx responses only. |
 |ignore_error| false | false | Validate and ignore result even if validation is error. So always return original data. |
 |parse_response_by_content_type| false | false | Parse response body to JSON only if Content-Type header is 'application/json'. When false, this always optimistically parses as JSON without checking for Content-Type header. |
+|strict| false | false | Puts the middleware into strict mode, meaning that response code and content type does not defined in the schema will be responded to with a 500 instead of application's status code. |
 
 No boolean option values:
 
--- a/lib/committee/middleware/response_validation.rb
+++ b/lib/committee/middleware/response_validation.rb
@@ -7,10 +7,7 @@ module Committee
 
       def initialize(app, options = {})
         super
-
-        unless options[:strict].nil?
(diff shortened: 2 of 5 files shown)


nicolasiensen:
Hey @ota42y, what is missing to merge this PR? I would be glad to help.

ota42y:
Sorry, there's nothing problem with this PR.
We have not changed master towards 5.0.0 yet, so we can't merge this PR.

Since 5.0.0 is a major version upgrade, we would like to include several breaking changes.
And we are thinking of making a version 4.99.0 that outputs various warnings to users won't have trouble when updating.
In order to do this, we need to decide what changes we want to make in 5.0.0.

committee use openapi_parser gem for validating OpenAPI 3 and I would like to release a major version upgrade (1.0.0) as the same time.
https://github.com/ota42y/openapi_parser

So we should do these step

1. release openapi_parser 1.0.0
2. use it in commitee 5.0.0 and add breaking changes
3. release 4.99.0 for comfortable upgrade
4. release 5.0.0 

In openapi_parser 1.0.0, drop old ruby version.
So we would like to add types and this is not done.
(This is because I add types by manually in many functions)
https://github.com/ota42y/openapi_parser/pull/118

But it's taking too long..., and it's not a functional addition, I think I'll release to using more `Any` type 🤔

nicolasiensen:
Regarding the `openapi_parser`, do you think it is a good idea to bump the Ruby version and add types in one go? Wouldn't it be more manageable if we bump Ruby first and add types in another version?

Besides bumping Ruby to 3, is there any other breaking change you plan to introduce to the `openapi_parser` 1.0?

ota42y:
My concern was that changing the defined form later would have to be treated as a breaking changes (and we need to update major version).
However, when I examined the version control in typescript, it was the management that the types can be changed frequently even in the same minor version.
As a result, I no longer have to worry about adding a type first and changing it later, so I'm going to release it with the current types.

nicolasiensen:
That makes sense. Let me know if there is anything I could help with, ok?

ota42y:
sorry, master branch's version changed 5.0.0, so I merge this PR

ota42y:
I released 5.0.0 with this changes!
https://rubygems.org/gems/committee/versions/5.0.0

PR #331, opened 2021-08-20
Allow `HEAD` method in v3

Pritilender:
I've followed [this commit](https://github.com/interagent/committee/commit/c0fe6d47c11a83e52909e92944aa0e0a951b97dd#diff-dd6da75ddbbb89e04e9c6d6320d52a08fd7232b16a167c9b256a437ab06ee8fb) which added support for `DELETE` method.

Please take a look.

### DIFF
--- a/test/data/openapi3/normal.yaml
+++ b/test/data/openapi3/normal.yaml
@@ -113,6 +113,26 @@ paths:
                 properties:
                   integer:
                     type: integer
+      parameters:
+        - name: limit
+          in: query
+          description: maximum number of characters
+          required: false
+          schema:
+            type: integer
+            format: int32
+    options:
+      description: preflight request
+      responses:
+        '200':
+          description: success
+          content:
+            application/json:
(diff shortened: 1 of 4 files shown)


test/data/openapi3/normal.yaml
```diff
@@ -113,6 +113,26 @@ paths:
...
                     type: integer
+      parameters:
+        - name: limit
+          in: query
+          description: maximum number of characters
+          required: false
+          schema:
+            type: integer
+            format: int32
+    options:
```
Pritilender, line 124: `options` is also supported in v3, as well as `trace`, so maybe it would make sense to add support for it as well? Though not sure which method could be used then for validating the `else` case for the `case-when`.

ota42y (approved):

ota42y:
I released 5.0.0 with this changes!
https://rubygems.org/gems/committee/versions/5.0.0

PR #353, opened 2022-02-17
Accept callable :error_class in middleware options.

iliabylich:
This fix allows autoloadable constants to be passed as `:error_class`.

Rails 7 doesn't "enable" constants autoload while running application.rb and initializers,
and so during middlewares initialization it's simply impossible to access any constant
from the `app` directory. By wrapping it with lambda we can lazy-load error_class.


Flixt:
Any particular reason why this PR was closed?

Issue #359, opened 2022-05-24
Rails deprecation warning for using `ActionDispatch::(Request|Response)#content_type`


ohbarye:
Hi 👋 

## Problem

I upgraded my Rails app and got the following deprecation warning from my code calling `assert_schema_conform`.

```
DEPRECATION WARNING: Rails 7.1 will return Content-Type header without modification. If you want just the MIME type, please use `#media_type` instead.
```

Maybe, the warning could happen from not only `Committee::SchemaValidator.request_media_type` but also `Committee::SchemaValidator::HyperSchema::ResponseValidator#response_media_type`.

## Reproduction

Call `assert_schema_conform` in Rails 7.0.0 or above.

## Expect

DEPRECATION WARNING doesn't appear.

## Actual

DEPRECATION WARNING appears.

## Possible solution

According to [Rails Upgrade Guide](https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#actiondispatch-request-content-type-now-returned-content-type-header-as-it-is) and the warning itself, it should use `media_type` instead of `ActionDispatch::(Request|Response)#content_type` if it's possible.

## Environment

- Rails 7.0.3

ohbarye:
Note: I found the warning appears while `ActionDispatch::Request.return_only_media_type_on_content_type` is true in the 6.1 defaults. After I upgraded the `load_defaults` from 6.1 to 7.0, the `return_only_media_type_on_content_type` defaults to `false` and the warning disappeared.

PR #360, opened 2022-05-24
Use `media_type` instead of `content_type` to suppress content-type warning

ohbarye:
## Issue

closes https://github.com/interagent/committee/issues/359

## Change

As [deprecation warning](https://github.com/rails/rails/blob/40272988cec849c734ec02cfa89ff830ae019d9e/actionpack/lib/action_dispatch/http/mime_negotiation.rb#L38-L41) below says, if just MIME type is necessary, it should use `media_type`.

```
DEPRECATION WARNING: Rails 7.1 will return Content-Type header without modification. If you want just the MIME type, please use `#media_type` instead.
```

So this pull request just replaces `content_type` with `media_type`.

`ActionDispatch::(Request|Response)#media_type` seem to exist since Rails 3.x ([ref](https://apidock.com/rails/ActionDispatch/Request/media_type)), I believe it's okay to just replace.

### DIFF
--- a/lib/committee/schema_validator.rb
+++ b/lib/committee/schema_validator.rb
@@ -4,7 +4,7 @@ module Committee
   module SchemaValidator
     class << self
       def request_media_type(request)
-        request.content_type.to_s.split(";").first.to_s
+        request.media_type.to_s.split(";").first.to_s
       end
 
       # @param [String] prefix
--- a/lib/committee/schema_validator/hyper_schema/response_validator.rb
+++ b/lib/committee/schema_validator/hyper_schema/response_validator.rb
@@ -48,7 +48,7 @@ module Committee
         private
 
         def response_media_type(response)
-          response.content_type.to_s.split(";").first.to_s
+          response.media_type.to_s.split(";").first.to_s
         end
 
         def check_content_type!(response)


r7kamura:
committee depends on rack, not rails, and its gemspec indicates that committee supports rack 1.5 or higher. However, there appears to be no `Rack::Response#media_type` in rack 1.5.

Another point I was wondering, since `media_type` returns only the MIME-Type part, the `.to_s.split(";").first` part is no longer needed and should be removed at the same time with this change?

ohbarye:
Thanks for pointing them out. To sum up, do you suggest changing like below? Or, should committee not have Rails related adjustment since it doesn't depend on Rails itself?

```ruby
# lib/committee/schema_validator.rb
def request_media_type(request)
  request.media_type # just call media_type since Rack::Request can respond even in rack v1.5.0.
end
```

```ruby
# lib/committee/schema_validator/hyper_schema/response_validator.rb
def response_media_type(response)
  if response.respond_to?(:media_type)
    response.media_type 
  else
    response.content_type.to_s.split(";").first.to_s # for rack compatibility. In rack v 1.5.0, Rack::Response doesn't have media_type
  end
end
```

---

BTW, I didn't know the warning appears only during transition period:  https://github.com/interagent/committee/issues/359#issuecomment-1166174534

r7kamura:
Yes, the example above is so good I think 👍

ohbarye:
Thanks for your advice, I've fixed those points.

ohbarye:
thx for your review too!

PR #367, opened 2022-12-17
Update README.md

shuntagami:
## What
SSIA

## Why
I ran into the error `RuntimeError: please set object like 'last_request'` when I tried this code.

## Review Perspective
Tell me If there are same document to be replaced like here, I'll fix it!


ota42y:
Yes, this is bug 😢  Thanks!!!

Issue #362, opened 2022-08-17
Middleware deprecation warnings are confusing


acceptto-nhumble:
If you use `Committee::Middleware::RequestValidation` you get a deprecation warning telling you to "please set parse_response_by_content_type = false because we'll change default value in next major version." which doesn't make sense in this context. That's not even listed as an option for that middleware. I think you're using the same options validation logic for all the middleware, which leads to warnings that make no sense. I can silence the warning by setting that value, but does it make sense to tell users to set an option that has no bearing on the middleware they're using?

It would be best if the options validation was aware of which middleware it was validating for so that it could give the correct responses. It could even warn about using options intended for other middleware. If you need any more information, please let me know and I am happy to help.

madleech:
Agreed, setting options that aren't actually options looks wrong:
```ruby
  use Committee::Middleware::RequestValidation, schema_path: SCHEMA_PATH, strict: true, query_hash_key: 'rack.request.query_hash', parse_response_by_content_type: true
  use Committee::Middleware::ResponseValidation, schema_path: SCHEMA_PATH, query_hash_key: 'rack.request.query_hash', parse_response_by_content_type: true
```

ota42y:
OK, we don't need set `parse_response_by_content_type` on RequestMiddleware so I remove deprecated message in v4.99.1
https://github.com/interagent/committee/pull/383

https://rubygems.org/gems/committee/versions/4.99.1

PR #383, opened 2023-01-28
supress parse_response_by_content_type deprecated message for request validation

ota42y:
When we use RequestValidation, parse_response_by_content_type doesn't support but show deprecated message.

close https://github.com/interagent/committee/issues/362

### DIFF
--- a/lib/committee/middleware/base.rb
+++ b/lib/committee/middleware/base.rb
@@ -13,7 +13,7 @@ module Committee
         @raise = options[:raise]
         @schema = self.class.get_schema(options)
 
-        @router = @schema.build_router(options)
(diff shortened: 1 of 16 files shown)


lib/committee/middleware/base.rb
```diff
@@ -13,7 +13,7 @@ def initialize(app, options={})
         @raise = options[:raise]
         @schema = self.class.get_schema(options)
 
-        @router = @schema.build_router(options)
+        @router = @schema.build_router(options, is_request: Committee::Middleware::RequestValidation == self.class)
```
ota42y, line 16: This code is bad code, we should divide option for RequestValidation and ResponseValidation.
But this code will be removed next major version and this commit doesn't use next version so its ok.

PR #396, opened 2023-07-04
Rename `parameter_overwite_by_rails_rule` to `parameter_overwrite_by_rails_rule`

r7kamura:
"overwite" is probably a typo for "overwrite" ?

- https://github.com/interagent/committee/pull/374

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- Rename `parameter_overwite_by_rails_rule` to `parameter_overwrite_by_rails_rule` #396
 
 ## [5.2.0] - 2024-05-04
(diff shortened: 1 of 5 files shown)


ydah (commented):
This option has already been released in v5.0.0, so it would be a good to change it while maintaining backwards compatibility.
It would be a good to proceed as follows in stages.

1. Add parameter_overwrite_by_rails_rule option to indicate to use it (or print a warning)
2. Release the above as a minor version update
3. Delete parameter_ overwrite _by_rails_rule during the next major version update

brandur:
Excellent point @ydah! It'd be great to fix this, and we should, but given that it was released already, we should be a little more careful about it.

@r7kamura Thank you for change! Sorry about the long delay here, but if you make the requested update above, we'll try to get it ushered in.

r7kamura:
OK, I've added one more commit to support the "overwite" one as well, so please check it out.

ydah (commented):
@r7kamura Thank you! Looks good!
I think this is a change that should be communicated to users, so could you add a description of the change in the CHANGELOG?

ydah (approved):
Thank you!

PR #400, opened 2024-01-12
Bump openapi_parser dependency to 2.0

aburgel:
v2.0 of openapi_parser has a [very useful fix to support nullable anyOf, allOf, oneOf types](https://github.com/ota42y/openapi_parser/pull/128). This is a common way to handle nullable references.

### DIFF
--- a/committee.gemspec
+++ b/committee.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |s|
   s.add_dependency "json_schema", "~> 0.14", ">= 0.14.3"
 
   s.add_dependency "rack", ">= 1.5"
-  s.add_dependency "openapi_parser", "~> 1.0"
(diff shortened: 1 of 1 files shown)


brandur:
@aburgel Now that #401 is in, can you rebase this and we can double check that everything checks out in CI? Thanks!

aburgel:
@brandur all pass except ruby 2.6. It looks like this new version of openapi_parser only supports 2.7 and up.

2.6 is pretty old and many libraries no longer supports. Rails dropped it in 7.0. So I think dropping it is reasonable.

brandur:
@aburgel Thanks. Yes I agree on 2.6 — it's been end of life'd for almost two years now, so it's okay to drop it. Did so in #403.

brandur (approved):

PR #401, opened 2024-01-12
Run CI workflow on pull requests

aburgel:
I noticed that CI only runs when you push to master, not on pull requests. I also took the liberty of removing the macos runner, it seemed a bit unnecessary since there isn't anything platform specific in this library (I can add it back in if I'm missing something).


brandur:
Thanks! And yes, the Mac OS runner seems unnecessary these days.

PR #403, opened 2024-01-14
Drop support for Ruby 2.6

brandur:
Drop support for Ruby 2.6. 2.6 has been out of support for almost two
years now (Mar 2022), and is causing a few issues with upgrading
dependencies [1], so it's probably okay at this point for Committee to
drop support for it.

[1] https://github.com/interagent/committee/pull/400


brandur:
(I think this is going to be a relatively uncontentious change, so going to go ahead and self-approve it.)

PR #404, opened 2024-01-23
Change to use Ruby 2.7 in `required_ruby_version` and `TargetRubyVersion`

ydah:
Follow up: https://github.com/interagent/committee/pull/403


brandur (approved):
Works for me. Thanks!

PR #405, opened 2024-01-23
Add base64 to gemspec

ydah:
base64 is loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. https://github.com/interagent/committee/actions/runs/7624721002/job/20767482497?pr=404#step:4:5

PR #406, opened 2024-01-23
Organize and update rubocop

ydah:
RuboCop was fixed to an older version due to supporting Ruby 2.4 at that time. However, it currently supports versions 2.7 and above, so the version constraint has been removed. Additionally, unused and unnecessary extensions were removed, suppressing the following warnings.

```
❯ bundle exec rubocop
Inspecting 81 files
.................................................................................

81 files inspected, no offenses detected

The following RuboCop extension libraries are installed but not loaded in config:
  * rubocop-minitest
  * rubocop-rake
```

base64 is loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. 
Although rubocop so far depends on base64, rubocop has not been added to the runtime dependency so far. So, rubocop will not work on Ruby heads with rubocop prior to [v1.56.0](https://github.com/rubocop/rubocop/releases/tag/v1.56.0). So it is a good to keep the version up to date.

https://github.com/interagent/committee/actions/runs/7624721002/job/20767482497?pr=404#step:4:5

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,6 +6,7 @@ AllCops:
   # RuboCop has a bunch of cops enabled by default. This setting tells RuboCop
   # to ignore them, so only the ones explicitly set in this file are enabled.
   DisabledByDefault: true
+  SuggestExtensions: false
   Exclude:
     - 'examples/**/*'
     - 'vendor/**/*'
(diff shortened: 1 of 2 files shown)


ydah:
needs: https://github.com/interagent/committee/pull/405

brandur (approved):
Great, thanks! (And thanks for the detailed PR description.)

PR #409, opened 2024-01-24
Use `Object#stub` instead of rr

ydah:
Since the [last commit for rr is in 2022](https://github.com/rr/rr/commits/master/), and currently, only the `stub` feature of rr is being used in committee, I believe it is possible to replace this with [Object#stub](https://www.rubydoc.info/gems/minitest/5.3.0/Object:stub) provided by Minitest. In general, I think it's better to minimize dependencies. WDYT?


brandur (approved):
Interesting. Use of rr probably dates all the way back to when I wrote this originally — rr was kind of a trendy choice back in those days. Use of `Object#stub` seems like a good idea, especially given how few uses of mocking there are elsewhere.

PR #411, opened 2024-01-24
Fix Ruby warning

y-yagi:
This fixes the following warning.

```
interagent/committee/lib/committee/schema_validator/open_api_3.rb:90: warning: assigned but unused variable - is_form_params
```


brandur (approved):
Thanks!

PR #417, opened 2024-03-15
Add option to permit validation when schema and data are empty

chibicco:
## Background

In our project, there are scenarios where the API returns an empty response body with status code 200.
However, the current OpenAPI 2 schema validation does not allow an empty schema for the response body, causing validation errors in these cases.

Specifically, in cases such as the schema and controller example below, I expect the Committee::Middleware::ResponseValidation check to pass successfully. 

```yaml
"/pets/cat": {
  "get": {
    "description": "Returns pets which are cats",
    "operationId": "find pets which are cats",
    "responses": {
      "200": {
        "description": "empty schema"
      }
    }
  }
}
```

```ruby
class CatController
    # pets/cat
    get "/" do
        status 200
        present nil
    end
end
```

## Changes

To address this issue, I have added an option to allow an empty schema for the response body in OpenAPI 2.
This change enables the validation to pass when both the schema and response body are nil, accommodating the specific use case in our project.

1. If allow_blank_structures is set to true, the validation will pass if the schema is empty and the data is nil.
2. I have also modified the error message for the case where data is present and the schema does not exist.

### DIFF
--- a/lib/committee/schema_validator/hyper_schema/response_validator.rb
+++ b/lib/committee/schema_validator/hyper_schema/response_validator.rb
@@ -4,11 +4,12 @@ module Committee
   module SchemaValidator
     class HyperSchema
       class ResponseValidator
-        attr_reader :validate_success_only
+        attr_reader :validate_success_only, :permit_blank_structures
 
         def initialize(link, options = {})
           @link = link
           @validate_success_only = options[:validate_success_only]
+          @permit_blank_structures = options[:permit_blank_structures]
 
           @validator = JsonSchema::Validator.new(target_schema(link))
         end
@@ -39,9 +40,18 @@ module Committee
             return if data == nil
           end
 
-          if Committee::Middleware::ResponseValidation.validate?(status, validate_success_only) && !@validator.validate(data)
-            errors = JsonSchema::SchemaError.aggregate(@validator.errors).join("\n")
-            raise InvalidResponse, "Invalid response.\n\n#{errors}"
+          if permit_blank_structures && @link.is_a?(Committee::Drivers::OpenAPI2::Link) && !@link.target_schema
+            return if data.nil?
+          end
+
+          begin
+            if Committee::Middleware::ResponseValidation.validate?(status, validate_success_only) && !@validator.validate(data)
+              errors = JsonSchema::SchemaError.aggregate(@validator.errors).join("\n")
+              raise InvalidResponse, "Invalid response.\n\n#{errors}"
+            end
+          rescue => e
+            raise InvalidResponse, "Invalid response.\n\nschema is undefined" if /undefined method .all_of. for nil/ =~ e.message
+            raise e
           end
         end
 
--- a/lib/committee/schema_validator/hyper_schema/response_validator.rb
+++ b/lib/committee/schema_validator/hyper_schema/response_validator.rb
@@ -4,12 +4,12 @@ module Committee
   module SchemaValidator
     class HyperSchema
       class ResponseValidator
-        attr_reader :validate_success_only, :permit_blank_structures
+        attr_reader :allow_blank_structures, :validate_success_only
 
         def initialize(link, options = {})
           @link = link
           @validate_success_only = options[:validate_success_only]
-          @permit_blank_structures = options[:permit_blank_structures]
+          @allow_blank_structures = options[:allow_blank_structures]
 
           @validator = JsonSchema::Validator.new(target_schema(link))
         end
@@ -40,7 +40,7 @@ module Committee
             return if data == nil
           end
 
-          if permit_blank_structures && @link.is_a?(Committee::Drivers::OpenAPI2::Link) && !@link.target_schema
+          if allow_blank_structures && @link.is_a?(Committee::Drivers::OpenAPI2::Link) && !@link.target_schema
             return if data.nil?
           end
 
--- a/lib/committee/schema_validator/option.rb
+++ b/lib/committee/schema_validator/option.rb
@@ -4,7 +4,8 @@ module Committee
   module SchemaValidator
     class Option
       # Boolean Options
-      attr_reader :allow_form_params,
+      attr_reader :allow_blank_structures,
+                  :allow_form_params,
                   :allow_get_body,
                   :allow_query_params,
                   :check_content_type,
@@ -17,8 +18,7 @@ module Committee
                   :optimistic_json,
                   :validate_success_only,
                   :parse_response_by_content_type,
-                  :parameter_overwite_by_rails_rule,
-                  :permit_blank_structures
+                  :parameter_overwite_by_rails_rule
 
       # Non-boolean options:
       attr_reader :headers_key,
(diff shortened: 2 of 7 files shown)


lib/committee/schema_validator/hyper_schema/response_validator.rb
```diff
@@ -39,9 +40,18 @@ def call(status, headers, data)
...
+            return if data.nil?
+          end
+
+          begin
+            if Committee::Middleware::ResponseValidation.validate?(status, validate_success_only) && !@validator.validate(data)
+              errors = JsonSchema::SchemaError.aggregate(@validator.errors).join("\n")
+              raise InvalidResponse, "Invalid response.\n\n#{errors}"
+            end
+          rescue => e
+            raise InvalidResponse, "Invalid response.\n\nschema is undefined" if /undefined method .all_of. for nil/ =~ e.message
```
chibicco, line 53: **ruby 3.2.2**

```ruby
NoMethodError: undefined method `all_of' for nil:NilClass
```

**ruby 3.3.0**

```ruby
NoMethodError: undefined method `all_of' for nil
```

**ruby head**

```ruby
NoMethodError: undefined method 'all_of' for nil
```

chibicco:
@brandur @ota42y 

If you have a moment, I would appreciate it if you could review it.
I apologize for my poor English skills, so please bear with me.

brandur:
@chibicco Sorry for the delay. Do you still need this?

chibicco:
@brandur 
Thank you for response :)

Yes, I need this pull requests.
I have rebased on the latest master, If possible, please review.

chibicco:
Greetings. 

I have just noticed the following correction. 
https://github.com/interagent/committee/pull/420

If my PR is not appropriate, I would appreciate it if you could close it.

brandur:
@chibicco Let's hold on that a bit longer. We have someone offering to help with maintenance so we may yet be able to rescue the project.

lib/committee/schema_validator/option.rb
```diff
@@ -46,6 +47,7 @@ def initialize(options, schema, schema_type)
         @optimistic_json                  = options.fetch(:optimistic_json, false)
         @parse_response_by_content_type   = options.fetch(:parse_response_by_content_type, true)
         @parameter_overwite_by_rails_rule = options.fetch(:parameter_overwite_by_rails_rule, true)
+        @permit_blank_structures          = options.fetch(:permit_blank_structures, false)
```
ydah, line 50: nits) As for the name of the parameter, how about `allow_blank_structures` to match the name of the other parameter if you want to allow it?

https://github.com/interagent/committee/blob/d1488a8d11cd323753e42587c48b45285640494f/lib/committee/schema_validator/option.rb#L42-L43

ydah, line 50: Could you add a description of this parameter to the README?

https://github.com/interagent/committee?tab=readme-ov-file#configuration-options

test/data/openapi2/petstore-expanded.json
```diff
@@ -184,6 +184,17 @@
...
         }
       }
+    },
+    "/pets/cat": {
+      "get": {
+        "description": "Returns pets which are cats",
+        "operationId": "find pets which are cats",
+        "responses": {
+          "200": {
+            "description": "empty schema"
```
ydah, line 194: imo) 204 No Content I think the response body is empty as a specification. So why not make the test data follow the common use case?

```suggestion
        "responses": {
          "204": {
            "description": "The resource was deleted successfully."
```

Refs: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204#compatibility_notes

ydah (commented):
@chibicco I left a few comments. Could you check?

ydah:
memo) I noticed in writing the following comment that 204 responses and so on expect to describe a response without a body. In other words, I think the next major update will change this parameters default value.

https://github.com/interagent/committee/pull/417#discussion_r1596566077

lib/committee/schema_validator/option.rb
chibicco: Thank you.
I have renamed permit_blank_structures to allow_blank_structures.

ref: https://github.com/interagent/committee/pull/417/commits/6de758350e2a7d0b1a7a49720637451d177cf8a3

chibicco: I have added a note to the README : )

[memo]

[In OpenAPI 3, the following becomes nil](https://github.com/ota42y/openapi_parser/blob/a430896dc0daa3b817e37e5c69791cf1841fb45b/lib/openapi_parser/schemas/response.rb#L26), so the validation of empty response succeeds.
Therefore, this option is expected to be valid only for Hyper-schema and OpenAPI 2.

test/data/openapi2/petstore-expanded.json
chibicco: In my recent experiences, I've noticed the following scenarios:

1. status_code=200 && empty response
=> This PR addresses this case. While it's technically correct to use status_code=204, there are instances where APIs are operated with 200, and ideally, I'd like to accommodate this case as well.

2. status_code=204 && empty response
=> While it's being skipped [here](https://github.com/interagent/committee/blob/10a4ebc14566ad8de02380cf68412b5b68e67367/lib/committee/middleware/response_validation.rb#L39) for 204, it seemed like a better approach to modify it to skip within Committee::SchemaValidator::HyperSchema::ResponseValidator. If it's not an inconvenience, I'm considering addressing this in a separate PR. Could you please share your thoughts on this?

chibicco:
@ydah 

Thank you for your review!
I have added the necessary commits for the corrections and replied to the comments.
I kindly ask for your confirmation.

test/data/openapi2/petstore-expanded.json
ydah: Thank you for your detailed explanation!
I understood the intent and thought it make sense. So I agree with your opinion.

ydah (commented):
@chibicco Thank you for your response! I think it looks good 😀
Finally, could you add a description of this change in CHANGELOG.md?

chibicco:
@ydah I have updated the CHANGELOG.md :)
We appreciate your kind support.

ydah (approved):
Looks good! Thank you for your work.

cc/ @brandur

brandur (approved):

chibicco:
@brandur 
Thank you for merging the PR.
Could you kindly release the latest master at a time that is convenient for you?

brandur:
@chibicco Yes! I believe @ydah is waiting for one more change on review feedback over on #396, after which he'll cut a release and get this shipped.

chibicco:
I see, I understand !

ydah:
@chibicco Thank you for waiting. Since v5.3.0 has been released, so your patch has been released! Thank you for your great work!
https://rubygems.org/gems/committee/versions/5.3.0

chibicco:
@ydah 

I have confirmed the v5.3.0 release : )
Thank you!

Issue #303, opened 2021-04-08
Committee OpenAPI 3 locked to version 3.0.x


hedselu:
Hi!

First of all, thanks for great gem. I just found a problem when I upgrade OpenAPI version in document to 3.1.0. 
The current implementation has locked version to 3.0.x, and fails finding paths for this document.

Current behaviour:
- runs HyperSchema driver for version OpenAPI version 3.x.x

Expected behaviour:
- runs OpenAPI3 driver for version OpenAPI 3.x.x

Related committee file for this issue: https://github.com/interagent/committee/blob/v4.3.0/lib/committee/drivers.rb#L52
Latest version of OpenAPI changelog: https://github.com/OAI/OpenAPI-Specification/releases

ota42y:
Thank you for reporting!
However, committee doen't support OpenAPI 3.1.0 yet.
(I have not been able to verify that it works properly)

So we can't load OpenAPI 3.1.0 now sorry

hedselu:
@ota42y understandable, thank you for answering!

Let me know if I should try verifying support for 3.1.x or someone else might take care of this. Of course I'd be happy to contribute.

ota42y:
I haven't check the OpenAPI 3.1.0 specification yet, so I haven't decided how much modification is needed ... 🙇

hedselu:
@ota42y no problem, thanks for response! 🙏

ahx:
One of the main differences with 3.1 and committee that I have noticed is the `type` field in JSON Schema. In 3.0 (and legacy JSON Schema version) it only accepts a string like `type: integer`, but in 3.1 (and at least JSON Schema draft-7) it support lists as well like  `type: [integer, string]` or even `type: [integer, null]` (which replaces `nullable`).

https://github.com/davishmcclurg/json_schemer supports draft-7.

chrishough:
@ahx that type nullable is really nice

geemus:
I've been working on compiling notes on changes that will be needed. I think also that we can probably add a little at a time (ie we don't have to wait to support all of 3.1 before we support parts). It may still take a little while though and I'd certainly welcome help from others to keep moving it forward.

geemus:
Hey, I agree that 3.1 would be great to continue working toward. That said, I think this is at least mostly a duplicate of #352. So I'm going to close this in favor of consolidating discussion in one place. I'd welcome folks to continue adding their questions and thoughts over there.

PR #418, opened 2024-03-19
Raise error when loading OpenAPI 3.1 schema

ytjmt:
Hi! Thank you for maintaining this great gem.
It can be confusing when the HyperSchema driver is used while loading the OpenAPI 3.1 schema. I think it would be beneficial to raise an error until OpenAPI 3.1 is officially supported.

https://github.com/interagent/committee/issues/303

> I just found a problem when I upgrade OpenAPI version in document to 3.1.0.
The current implementation has locked version to 3.0.x, and fails finding paths for this document.
> Current behaviour:
> - runs HyperSchema driver for version OpenAPI version 3.x.x

### DIFF
--- a/lib/committee/drivers.rb
+++ b/lib/committee/drivers.rb
@@ -70,6 +70,10 @@ module Committee
         return Committee::Drivers::OpenAPI3::Driver.new.parse(openapi)
       end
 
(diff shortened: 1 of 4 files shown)


brandur:
Thanks @ytjmt! Couldn't hurt to be explicit about this.

PR #420, opened 2024-05-04
Add maintenance warning to README

brandur:
Given that I rarely look at Committee anymore and other maintainers have
dropped off the project as well, add a maintenance warning to the README
indicating that new users should look elsewhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 A collection of middleware to help build services with JSON Schema, OpenAPI 2, OpenAPI 3.
 
+**MAINTENANCE WARNING:** Committee is an old project that's currently under extremely minimal maintenance. New users should look for alternatives instead.
+
 ## Supported Ruby Versions
 
 Committee is tested on the following MRI versions:


ydah:
@brandur If you or your team thinks so I would like to maintain this project. Because I am a frequent user of this project.

brandur:
@ydah Thank you for the kind offer! You have quite the maintenance history, so we'd be very excited to have you on board!

Just given that it would seem like poor practice to immediately add anybody to the org, would you mind if we ran a short trial period where we took a look at a few open changes?

In particular, we have a request in #417 where an active Committee user is trying to get a change in. Do you think you could take a look at that pull request and leave a review on it?

After that, happy to add you to the org and then we can revert the README note I added here.

Thank you!

ydah:
> Just given that it would seem like poor practice to immediately add anybody to the org, would you mind if we ran a short trial period where we took a look at a few open changes?
> In particular, we have a request in https://github.com/interagent/committee/pull/417 where an active Committee user is trying to get a change in. Do you think you could take a look at that pull request and leave a review on it?

Thank you for considering it even though it was a sudden offer!
No problem. I left a comment about #417 .

brandur:
Hi @ydah, thanks for looking into that.

I've added you as a member in GitHub and owner for the gem on RubyGems.

When you have a chance, why don't you try:

1. When satisfied with changes, merging #396 and #417.
2. Follow the README instructions to cut  new release: https://github.com/interagent/committee?tab=readme-ov-file#release
3. Revert the changes in this PR so that the project no longer reads as unmaintained.

Thanks!

PR #425, opened 2024-06-12
Ensure body is rewound

tt:
This is a small change to always rewind the request body. `Rack::Request#POST` no longer does this in recent Rack versions. It was removed in rack/rack@42aff22f708123839ba706cbe659d108b47c40c7.

`Rack::RewindableInput::Middleware` is likely going to be required for most users of this library.

Committee requires more changes to be compatible with Rack 3.1.x. I'm adding a constraint so Bundler doesn't try to upgrade dependencies past what's supported.

### DIFF
--- a/committee.gemspec
+++ b/committee.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |s|
   s.add_dependency "openapi_parser", "~> 2.0"
 
   s.add_development_dependency "minitest", "~> 5.3"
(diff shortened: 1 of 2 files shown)


beanieboi (approved):

ydah (approved):
Thank you! Looks good!

ydah:
@tt Do you need a release including this PR change sooner? If yes, we will release immediately.

tt:
@ydah, that would be excellent. Thank you.

PR #427, opened 2024-09-02
invoke error_handler on Committee::NotFound

brandonc:
My rationale is that an exception is raised if `raise` is true, but there is no opportunity to handle this error differently if otherwise.


ydah (commented):
Looks good

ydah (approved):

PR #428, opened 2024-09-11
Check Content-Type header for OPTIONS requests

Flixt:
We encountered a problem when a client made an `OPTIONS` request, because committee does only validate the content type header for `POST`, `PUT` and `PATCH` requests.

The OpenAPI (3) schema for the endpoint looks like this:

```yaml
MySchema:
  type: object
  required:
  - document_id
  properties:
    document_id:
      type: integer
```

The client left out the `Content-Type` header and sent and empty request body which did not lead to an `Committee::InvalidRequest` error.


ydah (approved):
Looks good for me.

PR #433, opened 2024-11-26
Support multiple response schemas for OpenAPI 2

renatolond:
Hi!

I noticed while testing OpenAPI3 vs OpenAPI2 that if a response was missing in the OpenAPI definition, then the OpenAPI2 driver would fail silently.

It's not the case for OpenAPI 3, where it says that the status code one tried to validate is not defined.

I feel that this might be a breaking change (since this will raise an error, where before it was working without problems), but I still think it's worth checking with you if you think it's a change worth merging into committee.

Let me know if you this makes sense or if you'd like me to change something!

### DIFF
--- a/lib/committee/drivers/open_api_2/link.rb
+++ b/lib/committee/drivers/open_api_2/link.rb
@@ -23,12 +23,19 @@ module Committee
 
         # The link's output schema. i.e. How we validate an endpoint's response
         # data.
-        attr_accessor :target_schema
+        attr_accessor :target_schemas
 
         attr_accessor :header_schema
 
         def rel
           raise "Committee: rel not implemented for OpenAPI"
+        end
+
+        def target_schema
+          target_schemas[status_success] ||
+            target_schemas[200] ||
+            target_schemas[201] ||
+            target_schemas.values.first
         end
       end
     end
--- a/lib/committee/drivers/open_api_2/driver.rb
+++ b/lib/committee/drivers/open_api_2/driver.rb
@@ -91,18 +91,18 @@ module Committee
 
         def find_best_fit_response(link_data)
           if response_data = link_data["responses"]["200"] || response_data = link_data["responses"][200]
-            [200, response_data]
+            200
           elsif response_data = link_data["responses"]["201"] || response_data = link_data["responses"][201]
-            [201, response_data]
+            201
           else
             # Sort responses so that we can try to prefer any 3-digit status code.
             # If there are none, we'll just take anything from the list.
             ordered_responses = link_data["responses"].
               select { |k, v| k.to_s =~ /[0-9]{3}/ }
             if first = ordered_responses.first
-              [first[0].to_i, first[1]]
+              first[0].to_i
             else
-              [nil, nil]
+              nil
             end
           end
         end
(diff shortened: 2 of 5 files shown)


renatolond:
Hey @geemus, no rush, enjoy your holidays!

I'm using [grape](https://github.com/ruby-grape/grape)/[grape-swagger](https://github.com/ruby-grape/grape-swagger) and currently it generates the docs in OpenAPI2, there's a few discussions and PRs around going to v3, but they are not final yet. Most notably https://github.com/ruby-grape/grape-swagger/issues/603 and https://github.com/ruby-grape/grape-swagger/pull/928, I'd say.
I would like to move to OpenAPI 3, but still haven't had the time to dive into the spec to see if I can help in the effort :)

lib/committee/drivers/open_api_2/link.rb
```diff
@@ -23,13 +23,20 @@ class Link
...
         def rel
           raise "Committee: rel not implemented for OpenAPI"
         end
+
+        def target_schema
+          target_schemas[status_success] ||
+            target_schemas[200] ||
+            target_schemas[201] ||
+            target_schemas.values.first
+        end
```
renatolond, line 39: Since I already call `find_best_fit_response` in the driver, I'm not sure this is needed here. Maybe just calling `target_schemas[status_success]` would be enough?
This is mostly for compatibility with hyper schema anyway.

geemus:
@renatolond sorry for the delays, spent a bit of time today trying to read and think through this. Could you share an (hopefully fairly minimal) example schema where it fails? Just want to make sure I'm on the same page of the exact culprit, and I might try to experiment a little with specifics once I have a repro case as I'm not quite sure how best to approach this. Thanks.

renatolond:
Yeah, I think it should be easy to test this with the test that already exist on committee, if you apply:
```
diff --git a/test/schema_validator/hyper_schema/response_validator_test.rb b/test/schema_validator/hyper_schema/response_validator_test.rb
index dee312b..573f275 100644
--- a/test/schema_validator/hyper_schema/response_validator_test.rb
+++ b/test/schema_validator/hyper_schema/response_validator_test.rb
@@ -101,6 +101,11 @@ describe Committee::SchemaValidator::HyperSchema::ResponseValidator do
     call
   end
 
+  it "returns a 404 response, not described in the schema" do
+    @status = 404
+    call
+  end
+
   it "raises errors generated by json_schema" do
     @data.merge!("name" => "%@!")
     e = assert_raises(Committee::InvalidResponse) { call }
```
Without my change (on master), this test will pass.

In this branch, it will raise:
```
  1) Error:
Committee::SchemaValidator::HyperSchema::ResponseValidator#test_0012_returns a weird response:
Committee::InvalidResponse: Invalid response./apps/{(%23%2Fdefinitions%2Fapp%2Fdefinitions%2Fname)} status code 404 definition does not exist
    lib/committee/schema_validator/hyper_schema/response_validator.rb:56:in `call'
    test/schema_validator/hyper_schema/response_validator_test.rb:121:in `call'
    test/schema_validator/hyper_schema/response_validator_test.rb:106:in `block (2 levels) in <top (required)>'
```

geemus:
@renatolond Thanks, I started digging through this more today, but haven't reached concrete conclusions quite yet. It's a bit complicated in there, so I may need a little more time to ponder and digest.

geemus:
@renatolond Sorry for the delays, I'd looked at this a couple times to see if I could find another way and never really landed on a different approach. As the tests pass it seems worth giving this a shot more or less as-is. That said, it seems it no longer cleanly applies (again, sorry for delays). Could you take a peak at getting this updated to be compatible with main again and I'll try to get it in more quickly this time around?

renatolond:
@geemus Pushed a rebase commit, should be good now :)

lib/committee/drivers/open_api_2/link.rb
geemus: Yeah, I'm not sure what is best (since I'm somewhat new to this codebase as well). Seems probably ok, so we can give it a shot.

geemus:
@renatolond thanks, lets bring this in then. I'll just note that given that we aren't quite sure the broader impact/etc I may end up needing to revert this if it causes bigger issues that we can't otherwise easily resolve. Hopefully it doesn't come to that, but I just want to set expectations.

renatolond:
Makes sense, thanks for taking the time to look into this! 💯

PR #438, opened 2025-01-20
Replace pry and pry-byebug with debug in development dependencies

ydah:
I don't have a strong opinion, but in most cases I think it is better to use what is provided by Ruby org. https://github.com/ruby/debug

This will also fix the following CI failures.

https://github.com/interagent/committee/actions/runs/12712838787/job/35884117256#step:4:14
````
/home/runner/work/committee/committee/vendor/bundle/ruby/3.5.0+0/gems/pry-0.14.2/lib/pry/command_state.rb:3: warning: ostruct was loaded from the standard library, but is not part of the default gems starting from Ruby 3.5.0.
You can add ostruct to your Gemfile or gemspec to silence this warning.
Stopped processing SimpleCov as a previous error not related to SimpleCov has been detected
/home/runner/.rubies/ruby-head/lib/ruby/3.5.0+0/bundled_gems.rb:76:in 'Kernel.require': cannot load such file -- ostruct (LoadError)
```


geemus (approved):
LGTM, thanks!

PR #439, opened 2025-01-20
Update .ruby-version to be the latest tested version

ydah:
ref: https://github.com/interagent/committee/pull/414


geemus (approved):
LGTM, thanks!

PR #440, opened 2025-02-05
Parse JSON when lower-case content-type is given

mopp:
Thank you for your nice gem to make our application robust.
I would like to fix a small problem.

## Problem to be solved

committee doesn't parse the response when `content-type: application/json` header is given.
Then, the validation failed due to comparing an object and String.

```json
{
    "id":"invalid_response",
    "message":"#/paths/~1characters/get/responses/200/content/application~1json/schema expected object, but received String: \"{\\\"Otonokizaka\\\":[\\\"Honoka.Kousaka\\\"]}\""
}
```

## How to solve it

Parse also the response when `content-type: application/json`.

## References

- [Are HTTP headers case-sensitive? - Stack Overflow](https://stackoverflow.com/a/5259004)
  - > Each header field consists of a name followed by a colon (":") and the field value. Field names are case-insensitive.
- [rack/SPEC.rdoc at v3.1.9 · rack/rack](https://github.com/rack/rack/blob/v3.1.9/SPEC.rdoc#the-headers-)
  - > Header keys must not contain uppercase ASCII characters (A-Z)
- https://github.com/ota42y/openapi_parser/pull/178#issuecomment-2494440535
  - This comment also refers to the same problem.

### DIFF
--- a/test/middleware/response_validation_open_api_3_test.rb
+++ b/test/middleware/response_validation_open_api_3_test.rb
@@ -17,6 +17,22 @@ describe Committee::Middleware::ResponseValidation do
     assert_equal 200, last_response.status
   end
 
+  it "passes through a valid response with content-type (lower-case)" do
+    status = 200
+    headers = { "content-type" => "application/json" }
+    response = JSON.generate(CHARACTERS_RESPONSE)
+
+    @app = Rack::Builder.new {
+      use Committee::Middleware::ResponseValidation, {schema: open_api_3_schema}
+      run lambda { |_|
+        [status, headers, [response]]
+      }
+    }
+
+    get "/characters"
+    assert_equal 200, last_response.status
+  end
+
   it "passes through a invalid json" do
     @app = new_response_rack("not_json", {}, schema: open_api_3_schema)
 
--- a/lib/committee/schema_validator/hyper_schema.rb
+++ b/lib/committee/schema_validator/hyper_schema.rb
@@ -29,7 +29,8 @@ module Committee
         data = {}
         unless full_body.empty?
           parse_to_json = !validator_option.parse_response_by_content_type ||
-                          headers.fetch('Content-Type', nil)&.start_with?('application/json')
+                          headers.fetch('Content-Type', nil)&.start_with?('application/json') ||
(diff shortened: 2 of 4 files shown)


test/middleware/response_validation_open_api_3_test.rb
```diff
@@ -17,6 +17,22 @@ def app
...
+    status = 200
+    headers = { "content-type" => "application/json" }
+    response = JSON.generate(CHARACTERS_RESPONSE)
+
+    @app = Rack::Builder.new {
+      use Committee::Middleware::ResponseValidation, {schema: open_api_3_schema}
+      run lambda { |_|
+        [status, headers, [response]]
+      }
+    }
```
mopp, line 30: I avoid touching the method `new_response_rack` to keep this PR small.
https://github.com/interagent/committee/blob/8e3e43d3494bff5a1a15d28f4b33e5cc0208d7b3/test/middleware/response_validation_open_api_3_test.rb#L285-L295

The method will be refactored after this PR is merged.

geemus:
Hey, thanks for the detailed report and references. I think it might make sense to think about using an approach very similar to [what we did in openapi_parser](https://github.com/ota42y/openapi_parser/pull/178#issuecomment-2494440535). At least presently we only really care about the content-type header, so we probably don't need the full Rack headers stuff, but it would still be good to allow any variation of capital letters for content type (so lower, title, upper or whatever mix really).

So we could use a detect + casecmp to figure out the variation in use:

```ruby
    content_type_key = headers.keys.detect { |k| k.casecmp?('Content-Type') }
```

And then lookup the value underlying that:

```ruby
    headers.fetch(content_type_key, nil)&.start_with?('application/json')
```

I think that feels like a relatively minimal change and would resolve the issue. What do you think?

Beyond that it looks like there might be some lint issues for us to cleanup, but we can deal with that once we settle on a fix. Thanks!

mopp:
@geemus Thank you for reviewing ✨ 
Your way looks good and I followed them.

I pushed the changes separately. But I can squash them into a commit if you prefer.

geemus (approved):
lgtm, thanks!

geemus:
released in v5.5.1

Issue #441, opened 2025-02-05
JSON body interaction


renatolond:
Hey!

I have an issue that I think it's on committee's side, but I'm not sure.

So, I'm using [Grape](https://github.com/ruby-grape/grape) for my API, with [rodauth](https://github.com/jeremyevans/rodauth)+[rodauth-oauth](https://gitlab.com/os85/rodauth-oauth) for the auth part.

Here's the issue: with grape reading the body, committee works fine. It seems grape rewinds the body before and after reading it, so it's ready for committee:
https://github.com/ruby-grape/grape/blob/master/lib/grape/middleware/formatter.rb#L86-L92

However, Roda doesn't if rack is >=3. https://github.com/jeremyevans/roda/blob/master/lib/roda/plugins/json_parser.rb#L89-L93

Now, the issue is that since committee does not rewind ahead of reading, it gets an empty body and assumes the `lenght == 0` situation (and therefore, no params) and fails the validation: https://github.com/interagent/committee/blob/master/lib/committee/request_unpacker.rb#L81-L83

Adding a rewind before reading solves the issue, but I'm not sure it's the right way to go (I read the Rack spec but it's not clear to me if that rewind is meant to always work).

I thought I could take the grape route and rewind if the input responds to rewind, wdyt?

renatolond:
Hmm, that does seem to solve the lack of `.rewind`, but I don't have that issue here. I'm not sure if it's because of my combination of gems, but I have `.rewind`, my issue is more that the IO has already been used by roda and not rewinded before used by committee. 🤔

geemus:
@renatolond Yes, sorry if that was unclear. The problem is more about doing this in a way that would still work for cases where there isn't rewind. So not so much solving your problem as avoiding creating new problems by solving this one. Does that make more sense?

And furthermore, the reason roda doesn't rewind is to avoid breaking things for streams (as suggested above). This was a more recent change, based on changes to Rack which allow for streams in that context in the first place.

PR #442, opened 2025-02-05
Rewind the body before reading and add safety around rewind

renatolond:
This is a proposed solution to #441, just to illustrate the issue and the solution.


geemus:
Added some context for discussion on #441.

PR #443, opened 2025-03-27
Add option to allow all GET and non-GET params in validation

### LINKED ISSUE #412: `assert_schema_conform` always fails with required query parameters for requests non-GET requests
## Overview

The following PR tightens the validation of query parameters:
* https://github.com/interagent/committee/pull/349

In some cases, query parameters cannot be passed to committee. Test failed because the required query parameters were not provided. This is caused by the use of query parameters in non-GET requests. In previous versions, parameters and request bodies were not rigorously checked, so this was not a problem, but in newer versions of committee, rigorously checked.

```
  1) /groups PATCH /groups/:id returns ok
     Failure/Error: assert_schema_conform 200
     
     Committee::InvalidRequest:
       #/paths/~1groups~1{id}/patch missing required parameters: name
     # ./spec/requests/groups_spec.rb:53:in `block (3 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # OpenAPIParser::NotExistRequiredKey:
     #   #/paths/~1groups~1{id}/patch missing required parameters: name
     #   ./spec/requests/groups_spec.rb:53:in `block (3 levels) in <top (required)>'
```

## Reproduction  code

I create a reproduction code: https://github.com/ydah/committee-v5-issue
There are two simple endpoints for this project:
https://github.com/ydah/committee-v5-issue/blob/4e1271fbde675a254b25f8a7b8c2867d397f04d9/app/controllers/groups_controller.rb#L4-L18

```ruby
  # GET /groups
  def index
    @groups = Group.all

    render json: @groups, status: :ok
  end

  # PATCH/PUT /groups/1
  def update
    if @group.update(group_params)
      render json: @group, status: :ok
    else
      render json: @group.errors, status: :unprocessable_entity
    end
  end
```

Both define required query parameters.
https://github.com/ydah/committee-v5-issue/blob/4e1271fbde675a254b25f8a7b8c2867d397f04d9/swagger/swagger.yml#L3-L49

```yaml
paths:
  /groups:
    get:
      summary: Get all groups
      parameters:
        - in: query
          name: limit
          schema:
            type: integer
          required: true
          description: The number of items to return¥
# (snip)
  /groups/{id}:
    patch:
      summary: Update a group
      parameters:
        - in: path
          name: id
          schema:
            type: integer
          required: true
          description: The group id
        - in: query
          name: name
          schema:
            type: string
          required: true
          description: The group name¥
```

Both pass query parameters as well, but succeed for GET and fail for PATCH, or non-GET, requests.

https://github.com/ydah/committee-v5-issue/blob/4e1271fbde675a254b25f8a7b8c2867d397f04d9/spec/requests/groups_spec.rb#L29-L43

```ruby
  describe "GET /groups" do
    it "returns ok" do
      Group.create!
      get groups_url, { limit: 1 }, as: :json
      assert_schema_conform 200
    end
  end

  describe "PATCH /groups/:id" do
    it "returns ok" do
      group = Group.create!
      patch group_url(group), { name: "something" }, as: :json
      assert_schema_conform 200
    end
  end
```

## Survey

It is not possible to pass query parameters to committee for non-GET requests since before the update. Here is where query parameters are actually retrieved:

https://github.com/interagent/committee/blob/f0611dc40ae92d2b468a2553cceb98878da0dc44/lib/committee/request_unpacker.rb#L55-L57

Retrieved with [Rack:: Request:: Helpers#GET](https://www.rubydoc.info/gems/rack/Rack%2FRequest%2FHelpers:GET), where query parameters are empty for non-GET requests.

For example, adding the following monkey patch shows that query parameters are not available for non-GET requests:

```ruby
module Committee
  class RequestUnpacker
    def unpack_query_params(request)
      pp request.GET
      pp request.POST
      @allow_query_params ? self.class.indifferent_params(request.GET) : {}
    end
  end
end
```

When the above test is performed, the results are as follows:

```
❯ bundle exec rspec spec/requests/groups_spec.rb
{"limit"=>"1"} # <--- index request.GET
{} #  <--- index request.POST
.{} # <--- patch request.GET
{"name"=>"something"} # <--- patch request.POST
F
```

That is, schema validation for non-GET requests and endpoints with required query parameters will always appear to fail.

zzak:
Fixes #412

This adds config to allow including all parameters when validating schema for GET and non-GET requests, disabled by default.

I understand the desire to only support query params for GET requests, but this behavior was working before, and our application depends on that behavior. So rather than rewriting those APIs, I think it's reasonable to have an option to support the existing behavior for legacy applications.

/cc @ydah

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ This piece of middleware validates the parameters of incoming requests to make s
 |allow_form_params | true | true | Specifies that input can alternatively be specified as `application/x-www-form-urlencoded` parameters when possible. This won't work for more complex schema validations. |
 |allow_get_body | true | false | Allow GET request body, which merge to request parameter. See (#211) |
 |allow_query_params | true | true | Specifies that query string parameters will be taken into consideration when doing validation. |
+|allow_non_get_query_params | false | false | Allow GET and non-GET query string parameters to be taken into consideration when doing validation. |
 |check_content_type | true | true | Specifies that `Content-Type` should be verified according to JSON Hyper-schema or OpenAPI 3 definition. |
 |check_header | true | true | Check header data using JSON Hyper-schema or OpenAPI 3 definition. |
 |coerce_date_times | false | true | Convert the string with `"format": "date-time"` parameter to DateTime object. |
--- a/lib/committee/request_unpacker.rb
+++ b/lib/committee/request_unpacker.rb
@@ -21,10 +21,11 @@ module Committee
     end
 
     def initialize(options = {})
-      @allow_form_params  = options[:allow_form_params]
-      @allow_get_body     = options[:allow_get_body]
-      @allow_query_params = options[:allow_query_params]
-      @optimistic_json    = options[:optimistic_json]
+      @allow_form_params          = options[:allow_form_params]
+      @allow_get_body             = options[:allow_get_body]
+      @allow_query_params         = options[:allow_query_params]
+      @allow_non_get_query_params = options[:allow_non_get_query_params]
+      @optimistic_json            = options[:optimistic_json]
     end
 
     # return params and is_form_params
@@ -57,7 +58,15 @@ module Committee
     end
 
     def unpack_query_params(request)
-      @allow_query_params ? self.class.indifferent_params(request.GET) : {}
+      unless @allow_query_params
+        return {}
+      end
+
+      if @allow_non_get_query_params
+        self.class.indifferent_params(request.params)
+      else
+        self.class.indifferent_params(request.GET)
+      end
     end
 
     def unpack_headers(request)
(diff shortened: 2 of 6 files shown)


ydah (approved):
Thank you! I think this makes sense. @interagent/committee-maintainers WDYT?

geemus (approved):
lgtm, thanks!

geemus:
I'm not sure what the default should really be, to be honest, but this shouldn't break anything and fixes something for folks that opt-in, so it seems like a great improvement.

geemus:
@zzak thanks! Do you need a release?

zzak:
@geemus Thank you! No rush, whenever you get around to it :bow:

geemus:
@zzak sure. I'd just as soon do it now while it has my attention (and therefore before I forget). Released in 5.5.2. Thanks!

PR #444, opened 2025-04-05
Add option to allow empty date and date-time strings

zzak:
Depends on ota42y/openapi_parser#182.

I wasn't sure how to version guard this, but the tests pass using my branch. But this won't work without a version of openapi_parser which includes that option.

I chose to test the validation error message, but there weren't other examples like this in the code base so maybe unnecessary (testing the parser, not the tool). More just me making sure things were lined up properly.

### DIFF
--- a/lib/committee/schema_validator/open_api_3/operation_wrapper.rb
+++ b/lib/committee/schema_validator/open_api_3/operation_wrapper.rb
@@ -31,10 +31,12 @@ module Committee
         end
 
         # @param [Boolean] strict when not content_type or status code definition, raise error
-        def validate_response_params(status_code, headers, response_data, strict, check_header)
+        def validate_response_params(status_code, headers, response_data, strict, check_header, validator_options: {})
           response_body = OpenAPIParser::RequestOperation::ValidatableResponseBody.new(status_code, response_data, headers)
 
-          return request_operation.validate_response_body(response_body, response_validate_options(strict, check_header))
+          return request_operation.validate_response_body(
+            response_body,
+            response_validate_options(strict, check_header, validator_options: validator_options))
         rescue OpenAPIParser::OpenAPIError => e
           raise Committee::InvalidResponse.new(e.message, original_error: e)
         end
@@ -89,11 +91,15 @@ module Committee
 
         # @return [OpenAPIParser::SchemaValidator::Options]
         def build_openapi_parser_option(validator_option, coerce_value)
-          datetime_coerce_class = validator_option.coerce_date_times ? DateTime : nil
-          validate_header = validator_option.check_header
-          OpenAPIParser::SchemaValidator::Options.new(coerce_value: coerce_value,
-                                                      datetime_coerce_class: datetime_coerce_class,
-                                                      validate_header: validate_header)
+          parser_options = {
+            coerce_value: coerce_value,
+            datetime_coerce_class: validator_option.coerce_date_times ? DateTime : nil,
+            validate_header: validator_option.check_header,
+          }
+          if OpenAPIParser::SchemaValidator::Options.method_defined?(:allow_empty_date_and_datetime)
+            parser_options[:allow_empty_date_and_datetime] = validator_option.allow_empty_date_and_datetime
+          end
+          OpenAPIParser::SchemaValidator::Options.new(**parser_options)
         end
 
         def validate_get_request_params(path_params, query_params, headers, validator_option)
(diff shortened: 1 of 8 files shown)


test/middleware/request_validation_open_api_3_test.rb
```diff
@@ -244,6 +244,40 @@ def app
...
+  it "errors given an empty datetime string with allow_empty_date_and_datetime disabled" do
+    @app = new_rack_app(schema: open_api_3_schema)
+    params = { "datetime_string" => "" }
+    get "/string_params_coercer", params
+
+    assert_equal 400, last_response.status
+    assert_match(/\\"\\" is not conformant with date-time format/i, last_response.body)
+  end
+
+  it "passes given an empty datetime string with allow_empty_date_and_datetime disabled" do
```
geemus, line 256:
```suggestion
  it "passes given an empty datetime string with allow_empty_date_and_datetime enabled" do
```

test/middleware/request_validation_open_api_3_test.rb
```diff
@@ -244,6 +244,40 @@ def app
...
+  it "errors given an empty date string with allow_empty_date_and_datetime disabled" do
+    @app = new_rack_app(schema: open_api_3_schema)
+    params = { "date_string" => "" }
+    get "/string_params_coercer", params
+
+    assert_equal 400, last_response.status
+    assert_match(/\\"\\" is not conformant with date format/i, last_response.body)
+  end
+
+  it "passes given an empty date string with allow_empty_date_and_datetime disabled" do
```
geemus, line 273:
```suggestion
  it "passes given an empty date string with allow_empty_date_and_datetime enabled" do
```

lib/committee/schema_validator/open_api_3/operation_wrapper.rb
```diff
@@ -89,9 +89,11 @@ def build_openapi_parser_path_option(validator_option)
 
         # @return [OpenAPIParser::SchemaValidator::Options]
         def build_openapi_parser_option(validator_option, coerce_value)
+          allow_empty_date_and_datetime = validator_option.allow_empty_date_and_datetime
           datetime_coerce_class = validator_option.coerce_date_times ? DateTime : nil
           validate_header = validator_option.check_header
-          OpenAPIParser::SchemaValidator::Options.new(coerce_value: coerce_value,
+          OpenAPIParser::SchemaValidator::Options.new(allow_empty_date_and_datetime: allow_empty_date_and_datetime,
```
geemus, line 95: I think a good backwards compatible guard for this might be something like:
```
if OpenAPIParser::Config.instance_methods.include?(:allow_empty_date_and_datetime)
```

So if that is true, you know it recognizes the option and you pass it in. And if it isn't defined it would probably be good to output a warning that tells you it won't do anything and if you need/want it to that you would need to update versions or something.

What do you think?

geemus (commented):
Looks good overall, suggested a couple of fixes for the test names and a possible way to guard for backwards compatibility, let me know what you think. Thanks!

lib/committee/schema_validator/open_api_3/operation_wrapper.rb
zzak: Good idea, I went with this:

```
if OpenAPIParser::SchemaValidator::Options.method_defined?(:allow_empty_date_and_datetime)
```

And then splat the keyword arguments.

I'm not sure if maybe using `method_defined?` is better because it allocates one less array (I didn't benchmark) but for us only using this in CI it doesn't matter. Happy to change it though if you have opinions.

zzak:
Hang on, I think we also need this in the response validator too.

```
Committee::InvalidResponse:
   #/components/schemas/Component/properties/originated_at Value: "" is not conformant with date-time format
 # lib/committee/schema_validator/open_api_3/operation_wrapper.rb:39:in `rescue in validate_response_params'
 # lib/committee/schema_validator/open_api_3/operation_wrapper.rb:34:in `validate_response_params'
 # lib/committee/schema_validator/open_api_3/response_validator.rb:20:in `call'
 # lib/committee/schema_validator/open_api_3.rb:46:in `response_validate'
 # lib/committee/test/methods.rb:40:in `assert_response_schema_confirm'
 # lib/committee/test/methods.rb:8:in `assert_schema_conform'
```

I'm looking into it, but I think we may need to patch `openapi_parser` again :sob:

zzak:
Ok, I've tested this with ota42y/openapi_parser#183 and everything is working now.

lib/committee/schema_validator/open_api_3/operation_wrapper.rb
geemus: Yeah, `method_defined?` seems better. I was looking for that and didn't find it at the time, so grabbed something that would work (and hopefully get the point across). Thanks for finding the cleaner way.

geemus (approved):
lgtm, thanks!

geemus:
@zzak thanks again for working through that. Is that everything you needed to change? Do we just need a committee release now?

zzak:
Thank you!! :bow:
Our tests are all passing with branches so I assume release shouldn't break anything. I will update it in my day tomorrow and let you know if anything goes wrong.

geemus:
Sounds good.

Released v5.5.3

Just let me know if anything comes up.

geemus:
Awesome, glad to hear it!

PR #447, opened 2025-06-30
Supports streaming response validation via `streaming_content_parsers` option

moznion:
## Background

In the existing implementation, when the response body of streaming events such as Server-Sent Events passes through `Committee::Middleware::ResponseValidation`, the `response_validate` method from [OpenAPI 3](https://github.com/interagent/committee/blob/cf8c0f268ec0c67117461aca3c9cd2fca8549b03/lib/committee/schema_validator/open_api_3.rb#L24-L27) (or [HyperSchema](https://github.com/interagent/committee/blob/cf8c0f268ec0c67117461aca3c9cd2fca8549b03/lib/committee/schema_validator/hyper_schema.rb#L24-L27)) blocks until the entire body is read. As a result, it was not possible to return responses as streams.

As a workaround for this issue, the one of the solutions is to use a middleware wrapper like the below to completely skip validation for the corresponding endpoint.

```ruby
class ConditionalCommitteeResponseValidator
    @app = app
    @middleware = Committee::Middleware::ResponseValidation.new(app, options)
  end

  def call(env)
    path_info = env['PATH_INFO']

    return @app.call(env) if path_info == '/events/stream'

    @middleware.call(env)
  end
end
```

## Features Added by This Patch

This patch adds an option called `streaming_content_parsers` to `Committee::Middleware::ResponseValidation`. This option accepts a hash where the key is a Content-Type and the value is a Proc that receives and transforms the response body of that Content-Type.

Content-Types specified via this option bypass the normal path (which waits for the entire body to be read) and instead go through a path via `Rack::BodyProxy`. This allows the response to be streamed to the client without blocking, while still enabling validation to be performed on the transformed body via the Proc once the response body is closed.

### Usage Example

```ruby
use Committee::Middleware::ResponseValidation,
  schema_path: 'docs/schema.json',
  streaming_content_parsers: {
    'text/event-stream' => ->(body) { body },  # Pass through SSE as string
    'application/x-json-stream' => ->(body) { JSON.parse(body) }  # Parse custom JSON stream
  }
```

### Limitations

For limitations of this feature, please refer to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -335,6 +335,8 @@ use Committee::Middleware::ResponseValidation,
 
 This allows you to validate streaming response formats by providing custom parsers for specific content types. The parser lambda receives the complete response body as a string and should return the parsed data for validation.
 
+**Note:** This feature uses `Rack::BodyProxy` internally, which buffers the entire response body in memory before validation. This can consume significant memory for large streaming responses. Consider this limitation when validating large data streams.
+
 Limitation: For standard Content-Types, it is possible to intervene in the actual response when validation fails (e.g., by setting the HTTP status code to 500). However, for Content-Types specified via the `streaming_content_parsers` option, validation is performed only after the entire response body has been read—that is, after it has already been returned to the client—making it impossible to modify the response. In such cases, only `handle_exception` is called, and an error is raised if necessary.
 
 ## Using OpenAPI 3
--- a/README.md
+++ b/README.md
@@ -175,6 +175,7 @@ No boolean option values:
 |prefix| String | support | support | Mounts the middleware to respond at a configured prefix. |
 |error_class| StandardError | support | support | Specifies the class to use for formatting and outputting validation errors (defaults to `Committee::ValidationError`). |
 |error_handler| Proc Object | support | support | A proc which will be called when error occurs. Take an Error instance as first argument, and request.env as second argument. (e.g. `-> (ex, env) { Raven.capture_exception(ex, extra: { rack_env: env }) }`) |
+|streaming_content_parsers| Hash[String, Proc] | support | support | A hash mapping content types to custom parser lambdas for streaming responses (e.g. `{ 'text/event-stream' => ->(body) { body } }`). Used for validating streaming response formats like Server-Sent Events. |
 
 Given a simple Sinatra app that responds for an endpoint in an incomplete fashion:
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -319,6 +320,23 @@ Committee has few options which enable convert request data.
 By default committee save converted data to `committee.params` and rails does not read it.
 So we need to save converted value to `'action_dispatch.request.request_parameters'` because rails create parameter from this value.
 
+### Streaming Response Validation
```
geemus, line 323: If I recall correctly, I think using the BodyProxy would also result in higher memory usage (because the whole response will be held in memory). For individual requests this might be alright, but in aggregate it might really add up. We should probably at least mention this as something to be aware of. What do you think?

lib/committee/middleware/response_validation.rb
```diff
@@ -9,25 +9,38 @@ def initialize(app, options = {})
...
-          v.response_validate(status, headers, response, @strict) if v.link_exist? && self.class.validate?(status, validate_success_only)
+        streaming_content_parser = retrieve_streaming_content_parser(headers)
 
-        rescue Committee::InvalidResponse
-          handle_exception($!, request.env)
+        if streaming_content_parser
+          response = Rack::BodyProxy.new(response) do
+            begin
+              validate(request, status, headers, response, streaming_content_parser)
+            rescue => e
```
geemus, line 24: I think for consistency, it might be good for this to have similar rescues/etc as the original version? Although, maybe that's not relevant because this happens after the response (and converting to 500s isn't really helpful any more). Is that what you were thinking here?

README.md
moznion: Yes, we’re on the same page. It depends on the response body size, but it often eats a lot of memory.

> We should probably at least mention this as something to be aware of. What do you think?

That sounds helpful, and I agree with your suggestion.
On the other hand, should we consider supporting an option to discard the response (i.e., skip validation)?

lib/committee/middleware/response_validation.rb
moznion: But when I tried to unify the error handling, the error couldn’t propagate from the `Rack::BodyProxy` code block to the unified handler (i.e., existing error handling logic).
I’m considering extracting the error handling logic into a separate function, but I’m not confident about it.
So for now, we might need to keep the current error handling as is — I mean, they should probably remain separate.

geemus: Sounds reasonable.

geemus (approved):
LGTM, sorry it took me so long to get back to it, July ended up being surprisingly hectic.

moznion:
@geemus
I deeply appreciate your support!
Since it’s not urgent, please feel free to prioritize any scheduled release items.

geemus:
@moznion This has been released as part of v5.5.5. Thanks!

PR #453, opened 2025-09-30
feat(schema-validator): return to lenient status parsing for hyperschema

### LINKED ISSUE #452: Breaking Change in HyperSchema ResponseValidator Status Code Handling
# Breaking Change in HyperSchema ResponseValidator Status Code Handling

## Problem Description

PR #433 introduced a breaking change in the [Committee::SchemaValidator::HyperSchema::ResponseValidator](cci:2://file:///Users/viniljain/code/committee/lib/committee/schema_validator/hyper_schema/response_validator.rb:5:6-103:9) that causes validation failures due to overly strict status code matching.

## Root Cause

The [ResponseValidator](cci:2://file:///Users/viniljain/code/committee/lib/committee/schema_validator/hyper_schema/response_validator.rb:5:6-103:9) now creates validators based on exact status code expectations from [link.status_success](cci:1://file:///Users/viniljain/code/committee/lib/committee/drivers/hyper_schema/link.rb:51:8-60:11):

```ruby
# In ResponseValidator#initialize
@validators[link.status_success] = JsonSchema::Validator.new(target_schema(link))
```

This creates a rigid mapping:
- `rel="create"` → expects exactly `201`
- `rel="empty"` → expects exactly `202` 
- All others → expect exactly `200`

## The Issue

When the actual response status code doesn't match the expected [link.status_success](cci:1://file:///Users/viniljain/code/committee/lib/committee/drivers/hyper_schema/link.rb:51:8-60:11), validation fails with:

```ruby
raise InvalidResponse, "Invalid response.#{@link.href} status code #{status} definition does not exist" if @validators[status].nil?
```

This happens because `@validators[status]` is `nil` when the status doesn't match the single expected value.

## Impact

This breaks backward compatibility for applications where:

1. **Create operations** return `200` instead of `201`
2. **Async operations** return `200` or `204` instead of `202`
3. **Existing working code** suddenly fails validation after upgrading

## Example Failure Scenario

```ruby
# Schema defines rel="create" (expects 201)
# But application returns 200 for creates
# Result: ValidationError because @validators[200] is nil
```

### Option 1: Flexible Status Code Ranges
Accept multiple valid status codes per relation type:

```ruby
def acceptable_statuses(rel)
  case rel
  when "create" then [200, 201]
  when "empty" then [200, 202, 204]
  else [200]
  end
end
```

### Option 2: Fallback Validation
Try exact status first, then fall back to common success codes:

```ruby
validator = @validators[status] || @validators[200] || @validators.values.first
```

### Option 3: Configuration Flag
Add `strict_status_codes` option to allow users to choose validation strictness.

geemus:
- #433 added strict checking for OpenAPI2, which broke hyperschema
- this maintains strict checking for OpenAPI2 while reverting hyperschema behavior change

fixes #452

### DIFF
--- a/lib/committee/schema_validator/hyper_schema/response_validator.rb
+++ b/lib/committee/schema_validator/hyper_schema/response_validator.rb
@@ -11,13 +11,14 @@ module Committee
           @validate_success_only = options[:validate_success_only]
           @allow_blank_structures = options[:allow_blank_structures]
 
+          @validator = nil
           @validators = {}
           if link.is_a? Drivers::OpenAPI2::Link
             link.target_schemas.each do |status, schema|
               @validators[status] = JsonSchema::Validator.new(target_schema(link))
             end
           else
-            @validators[link.status_success] = JsonSchema::Validator.new(target_schema(link))
+            @validator = JsonSchema::Validator.new(target_schema(link))
           end
         end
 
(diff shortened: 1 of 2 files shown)


ydah (approved):
Thank you! Looks good.

PR #471, opened 2026-03-25
Clarify OpenAPI 3 coerce_form_params behavior

### LINKED ISSUE #429: Documentation misleading about coerce_form_params option
From the documentation it says that `coerce_form_params` is used to coerce POST data encoded into an `application/x-www-form-urlencoded` body. 

But by looking at the source it seems it's not the behavior: The option is passed in [build_openapi_parser_body_option](https://github.com/interagent/committee/blob/ba2286f03131cd1cc9594666088b654032ebe9cc/lib/committee/schema_validator/open_api_3/operation_wrapper.rb#L82) and then forwarded in [build_openapi_parser_option](https://github.com/interagent/committee/blob/ba2286f03131cd1cc9594666088b654032ebe9cc/lib/committee/schema_validator/open_api_3/operation_wrapper.rb#L94) to the `OpenAPIParser::SchemaValidator::Options` class. 

I do not see anywhere that using this option, it's only applied to POST data encoded in `application/x-www-form-urlencoded`, it seems it's also applied for a JSON request body for example. Am I missing something?

ydah:
Fixes #429

- clarify README wording for OpenAPI 3 coerce_form_params behavior
- add a regression test showing that the option coerces JSON request bodies
- update the OpenAPI 3 driver comment to match the implementation


geemus:
LGTM, definitely appreciate the docs being accurate.

In the future we might consider adding a new flag with a clearer/more obvious name that reflects the new usage/expectations and just keeping this flag for backwards compatibility. That being said, that might also be confusing as it wouldn't really reflect pre openapi 3 behavior. Maybe it should be two flags? I don't know, this seems like the right thing to do for now, but maybe we can revisit at some point.

geemus (approved):

PR #472, opened 2026-03-26
Support bracket-style OpenAPI 3 query params

### LINKED ISSUE #327: Request validation does not support nested query parameters
First, thank you so much for building an amazing library!  It has helped us a lot in development of our new API, since we started with a documentation-driven approach.  It allows us to ensure our implementation matches our spec.  Please forgive me if this is a duplicate but I couldn't find it in your issue history.

We have some query parameters for an index operation that are defined as "filters" with a nested key for the attribute being filtered on.  An example is `GET /api/v1/events?filter[slug]=%2Ftest`.  [Rack's Query Parser implementation](https://github.com/rack/rack/blob/8be612ab949cf2eba7f4231ef17052a68315f911/lib/rack/query_parser.rb) turns the above into this params hash:
```ruby
{
  "rack.request.query_hash"=>{
    "filter"=>{
      "slug"=>"/test"
    }
  }
}
```

And then we get the following validation error:

```
#/paths/~1events/get missing required parameters: filter[slug]
```

Since it is a Rack specific behavior I didn't think this issue belonged in the `openapi_parser` library.

Thanks again!

ydah:
Fixes: https://github.com/interagent/committee/issues/327

This fixes OpenAPI 3 request validation for query parameters declared with bracket notation, for example:

```yaml
- in: query
  name: filter[slug]
  required: true
  schema:
    type: string
```

Rack parses a request like `?filter[slug]=/test` into a nested hash:

```
{ "filter" => { "slug" => "/test" } }
```

Before this change, Committee validated against the literal parameter name only, so the request was rejected as:

```
missing required parameters: filter[slug]
```

This change normalizes Rack-style nested query hashes into bracket notation before OpenAPI 3 parameter validation/deserialization when the schema expects bracket-style names.

### DIFF
--- a/lib/committee/schema_validator/open_api_3/parameter_deserializer.rb
+++ b/lib/committee/schema_validator/open_api_3/parameter_deserializer.rb
@@ -52,6 +52,8 @@ module Committee
           # If no parameters are defined for this location, return raw params as-is
           return raw_params if params_for_location.empty?
 
+          raw_params = normalize_raw_params(raw_params, location, params_for_location)
+
           # Collect parameter names that will be deserialized
           # This includes both the parameter name and any properties (for exploded objects)
           deserialized_keys = Set.new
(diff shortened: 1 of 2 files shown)


geemus (approved):
Oops, sorry for the delay. I saw the draft, but missed it when you marked it ready for review.

lgtm

ydah:
Thank you!

PR #473, opened 2026-04-06
Increment Minitest assertion count in assert_schema_conform methods

ursm:
## Summary

- `assert_request_schema_confirm` and `assert_response_schema_confirm` did not call Minitest's `assert`, so they were not counted as assertions. This caused Minitest to emit "Test is missing assertions" warnings when these were the only assertions in a test.
- Added `assert true` after successful validation to increment the assertion counter, guarded by `respond_to?(:assertions)` so it is a no-op in non-Minitest environments (e.g. RSpec).

## Test plan

- [x] Existing tests pass (`test/test/methods_test.rb`, `test/test/methods_new_version_test.rb`)
- [ ] Verify in a Minitest project that the warning no longer appears

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/committee/test/methods.rb
+++ b/lib/committee/test/methods.rb
@@ -94,6 +94,9 @@ module Committee
 
       private
 
+      # assert_*_schema_confirm signal failure by raising, not via `assert`,
+      # so Minitest would report "Test is missing assertions" on success.
+      # Bump the counter explicitly; no-op outside Minitest (e.g. RSpec).
       def increment_assertion_count
         assert true if respond_to?(:assertions)
       end
--- a/lib/committee/test/methods.rb
+++ b/lib/committee/test/methods.rb
@@ -23,6 +23,8 @@ module Committee
             schema_validator.request_validate(request_object)
           end
         end
+
+        increment_assertion_count
       end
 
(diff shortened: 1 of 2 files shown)


lib/committee/test/methods.rb
```diff
@@ -90,6 +94,10 @@ def old_behavior
 
       private
 
+      def increment_assertion_count
```
geemus, line 100: Could you add a small explanatory comment to the code here for posterity? The PR did a good job of describing the problem, I just worry that might get lost in the future and lead to confusion and/or regression.

In a similar vein, it would be great to have a test around this to prevent regressions as well.

Just let me know if you have questions or would like to discuss.

geemus (changes requested):
LGTM, thanks.

lib/committee/test/methods.rb
ursm: Thanks for the feedback! Added an explanatory comment on `increment_assertion_count` and regression tests for both `assert_request_schema_confirm` and `assert_response_schema_confirm` in 78e643c.

geemus (approved):
lgtm

ydah:
@ursm - released in v5.6.3

Issue #477, opened 2026-06-25
Feature request: opt-in strict validation when response Content-Type is not declared in spec


whoward:
**Problem**
When `Committee::Middleware::ResponseValidation` processes a response whose Content-Type doesn't match any media type declared in the spec's content block for that operation, it silently skips body schema validation entirely. The request passes as if it were valid.

This creates a false-confidence problem: a response carrying a content type that the spec doesn't declare at all will pass committee without any schema checking occurring. There's no way to know from the outside whether committee actually validated the response body or quietly gave up.

**Why it's hard to solve from outside the middleware**
The silent-skip is hardcoded deep in the validation flow. `openapi_parser`'s `select_media_type_from_content` returns `nil` when no match is found, and committee's response validation does return unless `media_type` — the skip. There is no hook, option, or callback to change this behaviour.

The only workarounds available from outside the middleware are:
- Monkey-patching `OpenAPIParser::Schemas::Response#select_media_type` directly — brittle and global
- Wrapping the app in a custom Rack layer that re-implements committee's path-matching logic just to replicate the content-type lookup - duplicates internals

**Proposed solution**
Add an opt-in option — `strict_response_content_type: true` (or whatever name feels right) - to `Committee::Middleware::ResponseValidation`. When enabled:
- If `select_media_type_from_content` returns `nil` for a response
- And the spec declares at least one media type for that operation's response (i.e. the `content` hash is non-empty)
- Raise `Committee::InvalidResponse` identifying the undeclared content type

The guard on `content` being non-empty matters: operations with no declared response body schema (bare 204 No Content, response objects with no content block) should still pass silently — the strictness only kicks in when the spec has opinions about what the content type should be and the actual response disagrees.

Defaulting to `false` preserves full backwards compatibility.

I'm happy to implement this if the approach sounds reasonable to you.

This seems to be similar to a previous issue #317 which implemented a similar approach but for response codes

geemus:
@whoward thanks for the detailed report, I think that makes sense. Just to be sure though, could you give a concrete example or two just to make sure we are on the same page? Otherwise, I think your proposal broadly sounds reasonable. I think it will be better to have this as a distinct option (vs changing the behavior of strict) for compatibility sake, at least for now. The proposed option name seems reasonable to me so far at least.

whoward:
Sure, here are a couple examples:

### Example 1 - content type declared, wrong type returned (should raise)

Spec declares:

```yaml
responses:
  '200':
    content:
      application/json:
[...]

PR #478, opened 2026-07-21
Add strict_response_content_type option to ResponseValidation middleware

whoward:
Committee silently skips response body schema validation when the response `Content-Type` doesn't match any key declared in the OpenAPI spec's content map for that operation. This means a server returning `application/vnd.api+json` against a spec that only declares `application/json` passes validation without error. The new opt-in flag `strict_response_content_type: false` (default `false` for backwards compatibility) raises `Committee::InvalidResponse` in this case. The guard is skipped for responses with no content map at all (e.g. bare 204s), which continue to pass silently.

The check is implemented as a pre-guard in `OperationWrapper#validate_response_params` before handing off to `openapi_parser`'s `validate_response_body`, since `openapi_parser` has no hook to express this policy without modifying the gem or repurposing the existing `strict:` flag. Status code lookup replicates `openapi_parser`'s wildcard matching (4XX, default) via a private `find_response_object_for_status` helper. `Content-Type` params (e.g. charset=utf-8) are stripped via `Rack::MediaType.type` before matching, consistent with how `validate_post_request_params` handles request body content types.

This work implements #477

### DIFF
--- a/lib/committee/middleware/options/response_validation.rb
+++ b/lib/committee/middleware/options/response_validation.rb
@@ -10,9 +10,10 @@ module Committee
         attr_reader :parse_response_by_content_type
         attr_reader :coerce_response_values
         attr_reader :streaming_content_parsers
+        attr_reader :strict_response_content_type
 
         # Default values
-        DEFAULTS = { strict: false, validate_success_only: true, parse_response_by_content_type: true, coerce_response_values: false }.freeze
+        DEFAULTS = { strict: false, validate_success_only: true, parse_response_by_content_type: true, coerce_response_values: false, strict_response_content_type: false }.freeze
 
         def initialize(options = {})
           super(options)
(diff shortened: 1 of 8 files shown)


geemus (commented):
Overall looks good to me, though it looks like our test coverage dropped below a threshold (you were unlucky in your timing maybe). That said, it would probably be good to unit test the find_response_object_for_status part as it is a bit involved. Could you look at fleshing that out a bit more (and if coverage is still insufficient, we can discuss other options/tweaks).

Thanks and apologies for my delay in reviewing.

whoward:
@geemus no worries! I believe I've fixed the coverage issue and have added unit tests covering `find_response_object_for_status` on it's own.  since this is a private method unit testing required the use of `send()` - hopefully that's not an issue.  Test coverage should now be 99.1%

geemus:
LGTM, thanks!

PR #479, opened 2026-09-01
Expose the matched operation on SchemaValidator::OpenAPI3

amogiska:
The OpenAPI 3 schema validator matches the request to an operation and holds it in `@operation_object`, but offers no public reader, so applications that want the matched operation have to use `instance_variable_get`.

One concrete use case: recording HTTP request metrics labeled by the matched `original_path` template, which keeps the label set bounded by the schema's paths instead of raw request paths (see ubicloud/ubicloud#6271, where a committee maintainer asked us to get a supported API here rather than read the ivar).

This adds a documented `attr_reader :operation_object` (returns the `OperationWrapper`, or nil when nothing matched) plus a test.

### DIFF
--- a/lib/committee/schema_validator/open_api_3.rb
+++ b/lib/committee/schema_validator/open_api_3.rb
@@ -50,6 +50,10 @@ module Committee
         !@operation_object.nil?
       end
 
+      # Expose the operation matched for the request, or nil if none matched
+      # @return [Committee::SchemaValidator::OpenAPI3::OperationWrapper, nil]
+      attr_reader :operation_object
+
       private
 
       attr_reader :validator_option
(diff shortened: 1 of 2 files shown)


geemus (approved):
lgtm, thanks!

iamKunalGupta:
Hi @geemus, is there any expected timeline on a new release of `committee`?

geemus:
@iamKunalGupta I don't really have a schedule per se, it's usually when something notable changes and/or somebody asks. I can probably do it early next week (I prefer earlier in the day and/or week when I'll be around/online more, just in case).

geemus:
@iamKunalGupta I just release 5.6.4, just let me know if you need anything further.

PR #481, opened 2026-09-08
Prevent path parameters from masking invalid query parameters

ydah:
OpenAPI 3 request validation merges path parameters into query parameters before validating them. When both locations declare the same name, the path value masks invalid or missing required query values and overwrites the query value after validation.

Pass only query parameters to `validate_request_parameter`, which validates queries and headers. Path parameters are already validated separately by `coerce_path_parameter`. Application-facing merged parameters retain their existing precedence.

Add GET and POST regression coverage for invalid values, missing required queries, independent coercion, and disabled query coercion. Update middleware assertions to verify that query values remain separate from path values.


geemus (approved):
lgtm, thanks!

PR #482, opened 2026-09-08
Validate OpenAPI 2 responses against status-specific schemas

ydah:
OpenAPI 2 response validation creates a validator for each response status, but initializes every validator with `link.target_schema`. Because that method selects a preferred success schema, responses for statuses such as 400 are also validated against the success schema.

Initialize each validator with the schema associated with its status in `link.target_schemas`.


geemus (approved):
lgtm, thanks!

PR #483, opened 2026-09-08
Preserve one-shot response bodies after validation

ydah:
Response validation enumerates the application response body to build the complete payload, then returns the same body object to the Rack server. Bodies backed by an IO, queue, or other one-shot enumerable are therefore empty when the server enumerates them again.

Buffer non-streaming response bodies selected for validation and use the buffered chunks for both schema validation and the returned Rack response. The original body is closed in an `ensure` block after it has been consumed. Responses that do not require validation and the existing `streaming_content_parsers` path retain their current behavior.

### DIFF
--- a/lib/committee/middleware/response_validation.rb
+++ b/lib/committee/middleware/response_validation.rb
@@ -28,8 +28,19 @@ module Committee
             end
           end
         else
+          validator = response_validator(request, status)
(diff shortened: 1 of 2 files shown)


geemus (approved):
lgtm, thanks!

PR #485, opened 2026-09-16
Validate each test request against its operation

ydah:
`Committee::Test::Methods#schema_validator` memoizes the validator created for the first request. OpenAPI 3 validators retain the operation matched during initialization, so subsequent requests in the same test instance are validated against the first request's operation.

Build the validator from the current request on every call while retaining the existing schema and router caches.


geemus (approved):
lgtm, thanks!

